PR #1, opened 2016-01-21
Webots documentation importer script

fabienrohrer:
Aim:

- [x] generate the Webots documentation into the extended markdown language, from the old Webots documentation system.
- [x] create a JS viewer page to show the result.
 
This first step is not about having a final solution, but only to explore this solution in depth.


fabienrohrer:
Ok, I have a working example, thanks to the `viewer.html` page on the `gh-pages` branch (using `showdownjs` for now):

http://omichel.github.io/webots-doc/viewer.html?book=guide&page=starting_chapter.md

fabienrohrer:
cf. https://github.com/omichel/webots-doc/blob/gh-pages/viewer.html

fabienrohrer:
Of course, several nodes are still not implemented or are buggy, and there is still no style sheet... But the mechanism is there..

fabienrohrer:
[Showdown](https://github.com/showdownjs/showdown) seems to be a very good candidate to be used from the viewer side:

- popular (used among others by stack overflow and google cloud platform) => long term support
- is working on the client side, only in javascript (no need to have a server running `php` or `nodes.js`). (it can be also implemented on the server side using nodes.js, but it's less useful for us..)
- simple to use. By comparison, I tried **quickly** [makdown.js](https://github.com/evilstreak/markdown-js) and [strapdownjs](http://strapdownjs.com/) unsuccessfully.
- it is doing the minimalistic features, but it is doing them well.
- HTML code without restrictions can be written into the markdown code. For example, add the following lines into the [demo editor](http://showdownjs.github.io/demo/):

``` html
<b>test</b>
<iframe width="560" height="315" src="https://www.youtube.com/embed/d6NJLFg1x9M" frameborder="0" allowfullscreen></iframe>
```

- it HTML is not sufficient, extensions can be written (e.g. to support x3d, maths, graphs, etc.). This is allowing us to master well our dependencies.
     - Extension doc: https://github.com/showdownjs/showdown/wiki/extensions
     - Extension example: https://github.com/showdownjs/youtube-extension

fabienrohrer:
The example shown in https://stackedit.io/editor is using a markdown library (not sure which one), and extend it to use other javascript library, such as:
- mathjax to display the mathematical functions: https://www.mathjax.org/
- js-sequence-diagrams and flowchart.js to display the charts: http://bramp.github.io/js-sequence-diagrams/ http://adrai.github.io/flowchart.js/

We can think about doing the same easily, or to choose another library to do so:
http://www.sitepoint.com/15-best-javascript-charting-libraries/
http://modeling-languages.com/javascript-drawing-libraries-diagrams/
(or more simply SVG).

fabienrohrer:
I think that the good working plan for this project is:
1. First milestone (this pull request):
    - to generate our documentation in the regular markdown language as-is and without new features and using static images (the `png` folder)
    - to create the corresponding viewer.html (in the `gh-branch` in order to see it through `github.io`) with a CSS for rendering it nicely in our website
    - to use this on our website
2. To generate the documentation of the `webots` develop branch into a new develop branch here
3. To drop our documentation on the `webots` branch
4. Improve it:
    - add new javascript libraries and extensions to support maths functions, graphs, x3d, etc.

omichel:
That seems good.

fabienrohrer:
Demo of the current state:

https://github.com/omichel/webots-doc/blob/feature-webots-doc-importer/reference/indexedfaceset.md
http://omichel.github.io/webots-doc/index.html?book=reference&page=indexedfaceset.md
https://github.com/omichel/webots-doc/blob/feature-webots-doc-importer/guide/using-the-thymio-ii-robot.md
http://omichel.github.io/webots-doc/index.html?book=guide&page=using-the-thymio-ii-robot.md

omichel:
Nice :smile:

fabienrohrer:
#### Reflexions on the exported HTML tags in the new documentation system

I'm writing this reflexion, in order to share the kind of issues I'm currently facing with this design.

By default `showdown` is exporting HTML tags without any layout/graphical attributes (without `id`, etc. but also without the `class` attribute).
This seems to be a very good constraint to keep. It's the aim of the CSS only to display it nicely. HTML and even less Markdown is supposed to deal with any layout/graphical information. At worst, the HTML tags should contain some `class` attributes.
Another constraint (but less important because we have a viewer) to keep if possible is to remain as close as possible of the original markdown language, otherwise, github, for example, won't be able to display the pages correctly (which is for now very convenient).

Note that I didn't found a clear way to add custom HTML attributes, except to create extensions, but it seems not perfectly related with my expectations: https://github.com/showdownjs/showdown/wiki/Cookbook:-Using-language-and-output-extensions-on-the-same-block

However, mechanisms should be found to display well our documentation related to this constraint.

For example, an image is defined like this in the markdown language:

```
![A simple IndexedFaceSet example](png/indexed_face_set.png)
```

`showdown` is exporting it like this:

```
<img src="png/thymio2.png" alt="Thymio II model in Webots">
```

We could use almost directly this to display our figures.

The issues are:
- our figures have other information (legend ~= figure's alt)
- to display a centered legend is problematic
    -  CSS is not able to display the image "alt" attribute (of course without js trick or so)
    - if I add a regular paragraph bellow, I don't have a way to filter it out of the other paragraphs from the CSS point of view, because it doesn't have a class
- to center the image
    - an intermediate tag would be required to do so, and showdown is not doing this by default.

I implemented for now an ugly intermediate solution: I choose for now is to write the figure like this in the markdown, in order to display it easily from the CSS:

```
<center>
![A simple IndexedFaceSet example](png/indexed_face_set.png)

#### A simple IndexedFaceSet example
</center>
```

But this is clearly not a good solution, because:
- the tag center is indicating layout information in the markdown
- the legend is duplicated
- the legend is put as a h4, which is quite hacky.

So, I think the best way is to deal with this kind of issue is to keep the very simplistic way to define a figure in the markdown:

```
<img src="png/thymio2.png" alt="Thymio II model in Webots">
```

.. or eventually, if it helps:

```
%figure%
#### Thymio II model in Webots
<img src="png/thymio2.png">
%end%
```

.. and to either create a new `showdown` rule, or to overwrite the `showdown` rule to generate the images, in order to export a more sofisticated HTML tag hierarchy (for example using the [HTML5 `figure` and `figcaption` tags](http://www.alsacreations.com/article/lire/1337-html5-elements-figure-et-figcaption.html) )

(Before implementing this, I have to solve the way to generate references or so, because it is linked with the dependency on the `showdown` library).

omichel:
The last option you propose for making use of the HTML5 figure tag seems the best because it will be easy to use CSS to display figures and caption in a various way (centered, caption in italics, etc.).

About references, can't we simply use the HTML A tag with hashtag option in href?

fabienrohrer:
> About references, can't we simply use the HTML A tag with hashtag option in href?

I just started to work on this, and this is the last main incertitude I have with the new system.

We can do obviously direct links using in markdown:
https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#links

But I'm asking if we can't do better, because it means that we have to maintain them hard coded. This means for example that if we change a filename, we have to make sure to change all the references too.

I wonder also how to deal with chapter/section/subsection/figures numbers. They become meaningless in the new system (replaced by slugs, TOCs, etc.).

We could think about adding scripts to check that the references are ok, and/or to generate a reference file which would be loaded into the viewer in order to create the links dynamically.

fabienrohrer:
And, it would be good to add anchors to each subsections to reference them quickly, as in the github wiki.

fabienrohrer:
A showdows extension is doing a part of the job: [showdown-toc](https://github.com/ravisorg/showdown-toc)

fabienrohrer:
New demo of the current state of the branch:

- http://omichel.github.io/webots-doc/index.html?book=guide&page=toc.md
- http://omichel.github.io/webots-doc/index.html?book=reference&page=toc.md

Any comment is welcome (except the bugs, because I still have to fix them :smile:  )

omichel:
It looks very nice! I like the menu on the left, althought I miss a little bit the visualization of the other subsections inside a section. I saw a couple of bug but won't report them for now :wink:

fabienrohrer:
> I miss a little bit the visualization of the other subsections inside a section.

I agree.

For now, I used the default jquery menu, without any modification:
https://jqueryui.com/menu/#default

It's a very good starting point, because we can achieve very different result, just by plaining with the CSS:
http://designscrazed.org/jquery-navigation-plugins-menus/

For example, displaying the second menu layer as we did before should be possible. I will check this.

fabienrohrer:
I finally found how to override a rule in showdown; an extension based on the regular expression rule to parse the image do the job:

https://github.com/showdownjs/showdown/blob/master/src/subParsers/images.js

``` javascript
showdown.extension("example", function() {
  'use strict';

  return [{
    type: 'lang',
    filter: function(text, converter, options) {
      text = text.replace(/!\[(.*?)]\s?\([ \t]*()<?(\S+?)>?(?: =([*\d]+[A-Za-z%]{0,4})x([*\d]+[A-Za-z%]{0,4}))?[ \t]*(?:(['"])(.*?)\6[ \t]*)?\)/g,
      function (wholeMatch, altText, linkId, url, width, height, m5, title) {
        // cf. writeImageTag() function
        console.log(altText);
        console.log(url);
        console.log(title);
      });
      return text;
    }
  }]
});
```

I'm still not perfectly safistyied by this approach, because the code will look like a fork of the `images.js` file above, and it's difficult to differentiate the images that we want to display as figures from the others.
I'm asking myself if adding a new rule wouldn't be better, because it would also work for other things like tables, like this:

```
%figure "A simple IndexedFaceSet example"
![A simple IndexedFaceSet example](png/indexed_face_set.png)
%end
```

fabienrohrer:
The solution explained just above about figures is now implemented, thanks to a showdown extension. It allows to display well the image and table figures.

Example:
http://omichel.github.io/webots-doc/index.html?book=guide&page=using-the-e-puck-robot.md

The only small drawback is that the github markdown is not reading well this extension, but I think it's the best solution anyway.
https://github.com/omichel/webots-doc/blob/feature-webots-doc-importer/guide/using-the-e-puck-robot.md

I understand well now how to write any showdown extension.

fabienrohrer:
status:
- the importer script is almost completed (only have to solve some small/misc issues) => the md files are close to the final version.
- the viewer is still to be improved.

fabienrohrer:
### New demo:

http://omichel.github.io/webots-doc/index.html?book=guide
http://omichel.github.io/webots-doc/index.html?book=reference

### Status:

I'm almost down :-)

- The markdown files are now completed. All the old documentation nodes are implemented. The rendering in Github is good looking.
- The viewer and the CSS is still to improve

### TODO list

- features
    - [ ] implement anchors on the viewer side
    - [ ] add an button to copy quickly the anchors into the clipboard (as in the github pages)
    - [ ] change the pages "on the fly" .i.e. without reloading the whole page
    - [ ] the level 2 of the menu should be visible
    - [ ] the icon of the "note", "handson" or so should be restored (probably thanks to a showdown extension)
    - [ ] implement a test script checking that:
        - [ ] all the references are used
        - [ ] all the references are pointing on existing resources
        - [ ] the `menu.md` file is referring to all the files
        - [ ] all the images are used
- known bugs (I didn't hunt them)
    - [ ] GET parameters should be implement in a more robust way
    - [ ] xref are not displaying correctly the kind tag ("figure" or so.)
- problematic things in the old documentation:
    - [ ] some legends of some figures are huge (mainly in the tutorial), this is problematic to reference them

omichel:
The demo doesn't seem to work for me...

fabienrohrer:
Very strange, I don't see any issue..
The page is not loading? Is there an issue in the javascript console?

stefaniapedrazzi:
It doesn't work for me as well.
No page is loaded. Here is the warning I get on Firefox:
```
SyntaxError: in strict mode code, functions may be declared only at top level or immediately within another function showdown-extensions.js:63:29
Error: Extension "wbfigure" could not be loaded. It was either not found or is not a valid extension.
```

omichel:
The menus are displayed properly, but the contents is not displayed at all and I get this error:
error

fabienrohrer:
Ok, I can reproduce it on Firefox, while it's working on Chrome. I'm investigating why..

fabienrohrer:
Ok, this is working again on firefox..

omichel:
I investigated and found that we can simply use the mod alias from apache2 (which seems to be enabled by default) and add a simple regex rule in `/etc/apache2/mods-enables/alias.conf`:
```
AliasMatch ^/test/(.*) /var/www/test.php
```
Then, a simple PHP script can display the page reference:
```php
<?php
  echo "URL is http://$_SERVER[HTTP_HOST]$_SERVER[SCRIPT_NAME]<br>\n";
  $i = strpos($_SERVER['SCRIPT_NAME'], '/', 1);
  $page = substr($_SERVER['SCRIPT_NAME'], $i + 1);
  echo "PAGE is $page<br>\n"
?>
```
I implemented a test on Cyberbotics web server and it seems to work fine:

https://www.cyberbotics.com/test/installing-chapter
https://www.cyberbotics.com/test/writting-a-c-robot-controller
https://www.cyberbotics.com/test/ceci-est-un-test

fabienrohrer:
I tried this morning to implement my first thought on this topic: to setup Apache2 with the rewrite module, but I didn't achieve a working setup. I think nevertheless it's also a good alternative.

https://wiki.apache.org/httpd/RewriteQueryString

An Apache rule like the following should do the job:

```
RewriteRule ^/docs/([^/]+)/([^/]+) http://omichel.github.io/webots-doc/index.html?book=$1&page=$2 [P]
```

.. but it's a complete redirection. With this technique, it's difficult to embed a page into another one.

fabienrohrer:
@omichel Could you create symlinks from:
FTP/repository/www/highlight/ -> http://www.cyberbotics.com/highlight/
FTP/repository/www/showdown/ -> http://www.cyberbotics.com/showdown/
.. as we do for jQuery and others?

omichel:
Done :smiley:

fabienrohrer:
Great, thanks

fabienrohrer:
I think it's time for a review :-)

MD files are visible on GitHub here:
https://github.com/omichel/webots-doc/blob/feature-webots-doc-importer/guide/guide.md
https://github.com/omichel/webots-doc/blob/feature-webots-doc-importer/reference/reference.md
https://github.com/omichel/webots-doc/blob/feature-webots-doc-importer/automobile/automobile.md
https://github.com/omichel/webots-doc/blob/feature-webots-doc-importer/darwin-op/darwin-op.md

The viewer is on the gh-pages branch:
https://github.com/omichel/webots-doc/tree/gh-pages

The results can be shown on this URL:
http://omichel.github.io/webots-doc/?book=guide
http://omichel.github.io/webots-doc/?book=reference
http://omichel.github.io/webots-doc/?book=automobile
http://omichel.github.io/webots-doc/?book=darwin-op

The tests can be launched with this command:
python -m unittest discover

Note:

I propose to deploy like this:
- The viewer remains on the gh-pages branch.
- The Cyberbotics site should use Apache/PHP to include the "http://omichel.github.io/webots-doc/?book=.*&page=.*" urls in it.

:+1:

DavidMansolino:
Just a few comment on the automobile doc result (as I am not sure how to review this PR I have checked the result for each page both for the md and the resulting page, only automobile for now):

  1. http://omichel.github.io/webots-doc/?book=automobile&page=driver-library.md: come paragraph/part of text change color when mouse is over it, this seems not normal (this happen in other pages too, seems a problem at the viewer level, as md file seems ok).
 2. https://github.com/omichel/webots-doc/blob/feature-webots-doc-importer/automobile/traffic-signs-and-lights.md The last figure should contain 4 images and not only 1. Same for http://omichel.github.io/webots-doc/?book=automobile&page=openstreetmap-importer.md, the first picture should contains 2 images, (and http://omichel.github.io/webots-doc/?book=automobile&page=village-winter.md)
 3. In the proto definition (e.g. http://omichel.github.io/webots-doc/?book=automobile&page=nature.md) the color is not homogeneous (proto name red vs normal, default value in red vs normal, MFString in bold and not the other, some field name in green some other not).

Of course I think not all of those bugs should mandatorily be fixed for the first version of the doc in md, it is just to let you know.
The resulting pages are very nice to see and read, the layout, color, etc. are very nice.
I also notice that the menu is not always working as expected and the page generation is way slower than with the current website.
I will now check the darwin-op doc to see if I can find other problems.

DavidMansolino:
Problem in the darwin-op doc:
  1. not linked with this PR but https://www.cyberbotics.com/darwin-op/motion-manager.php: reference to appendix is wrong.
  2. The link with external website seems not to be working: http://omichel.github.io/webots-doc/?book=darwin-op&page=motion-manager.md

fabienrohrer:
Thank you, I will fix all these bugs.

fabienrohrer:
I think now that it will be better to adopt the new documentation just after releasing Webots 8.4, because it will let us the time to familiarize/debug it.

fabienrohrer:
I fixed all the bugs reported by David. Except about the syntax highlight which is difficult to improve systematically.

> I also notice that the menu is not always working as expected

I'm not able to reproduce issues there. Could you help me to reproduce?
The Next and Previous buttons are only working to pass from chapter to chapter, or from section to section.


> and the page generation is way slower than with the current website.

Visually, I don't see particular performance issues.

Doing a more serious performance check using the "Chrome / Development Tools / Time line" tool, I get these loading time:

http://omichel.github.io/webots-doc/?book=darwin-op&page=vision-manager.md -> ~450 [ms]
https://www.cyberbotics.com/darwin-op/vision-manager.php -> ~350 [ms]

This performance difference can be explained by:
- the 2 ajax calls (to get the MD file and the menu),
- the showdown parsing (75 [ms] !)
- the highlight parsing

 So I think that the performance difference is acceptable for now.

stefaniapedrazzi:
I also notice that loading a page is sometimes slow and takes a couple of seconds.
Mainly the first time a page is requested but it is not systematic.

DavidMansolino:
Concerning the performances, it seems also that images take more time to load (not 100%sure).

> The Next and Previous buttons are only working to pass from chapter to chapter, or from section to section.

Those one seems to work fine. But when you click on one of the chapter menu, sometimes it open the chapter page and some other time it open the sub-menu, the behaviour is not always the same and not reproducible.

omichel:
> Doing a more serious performance check using the "Chrome / Development Tools / Time line" tool, I get these loading time:

Did you clear the cache to compare the loading times?

fabienrohrer:
The first load is indeed much slower, because of the loading of the JS libraries, and probably of the images, etc. But the same can be said about the cyberbotics current website.

stefaniapedrazzi:
From what I could experience the current cyberbotics website is visibly faster.

fabienrohrer:
I'm wondering why I cannot see this difference.. Maybe I don't remove correctly the cache..

fabienrohrer:
On Firefox, the loading time is indeed bigger than on Chrome.

DavidMansolino:
> From what I could experience the current cyberbotics website is visibly faster.

Yes, I had the same feeling, the current cyberbotics pages seems to load 'instantaneously' but on this system you kind of 'see' that the page is loading.

fabienrohrer:
In fact, it's also what's going on internally => Once the index.html is loaded, then an ajax request grabs the MD file, and converts it to HTML before loading it in the index.html page.
It's much more computer intensive than simply loading a page which is already in HTML.

fabienrohrer:
I can propose to implement the on-the-fly load of the pages (not too difficult I guess).
Like this, the navigating from one page to another would be done without reloading all the page, but only the concerned div. In this case, we could expect to have even better performance than the current web site (the JS libraries should not be downloaded and reloaded). But for sure, the problems will remain on the first documentation page load.

omichel:
But in any case, these JS libraries should be cached by the browser and therefore should not downloaded again and again, isn'it?

fabienrohrer:
Yes, I would expect this. I'm not at 100% sure how the browser cache is working.

omichel:
If this on-the-fly load is easy to implement, it is certainly worth trying to see whether it improves speed or not.

fabienrohrer:
Yes, I started to implement this.
It would also probably facilitate the integration of the page in our website: no need to convert the hyperlinks..

fabienrohrer:
Could you tell me if the speed has been increased:

http://omichel.github.io/webots-doc/?book=guide
http://omichel.github.io/webots-doc/?book=reference
http://omichel.github.io/webots-doc/?book=automobile
http://omichel.github.io/webots-doc/?book=darwin-op

One remaining issue is that the MD files are read from the github raw links rather than in the github.io interface which is probably optimized.
In any case, I think that on this repo, we should drop the `master` branch, and work on the `gh-pages` branch as it was the master branch, like this, we can take the advantage of the github.io interface.

DavidMansolino:
Yes, the speed seems to have considerably increased :smiley:

stefaniapedrazzi:
Now it seems faster, but it is strange to see the page layout changing while loading.

I also noticed that the menu item is loaded only if you click on the text, but not if the item line is selected but you click on the right. This is quite annoying because given that the item is selected it is not easy to understand why it doesn't work.

fabienrohrer:
It's great to read the performance has been improved.

> it is strange to see the page layout changing while loading.

Yes, I played with the jquery animations, I admit this is not standard at all :-)
I can switch back to the not-animated version by changing one line..


> I also noticed that the menu item is loaded only if you click on the text.

Yes, I should indeed put the hyperlink on the div rather than on the text.

fabienrohrer:
I have to reopen this PR with gh-pages branch as target.

PR #2, opened 2016-03-09
New Webots documentation system

fabienrohrer:
Aim:

- [x] generate the Webots documentation into the extended markdown language, from the old Webots documentation system.
- [x] create a JS viewer page to show the result.
 
This first step is not about having a final solution, but only to explore this solution in depth.


fabienrohrer:
Not working as I would like, because `gh-pages` is currently empty, so I cannot see the result on github.io. I have to put all the code into the `gh-pages` branch directly.

PR #9, opened 2016-04-14
Feature math formulas

### LINKED ISSUE #7: Mathematical functions
For now, the formula sources are still into the TEX format in the `webots/src/doc/formulas/` repository.

A better integration seems obviously required.

Several Javascript are doing mathematical text to HTML formula view conversion very well including:
- [MathJax](https://www.mathjax.org/)
- [AsciiMath](http://asciimath.org/)

MathJax is supposed to support TEX formulas, amongs other, which seems to be the shortest way to import our formulas.

Some code snippet are existing to integrate this into showdown:
https://gist.github.com/xentec/6588519
https://gist.github.com/npisenti/3028983

fabienrohrer:
Attempt to fix #7

- [x] Replace images with TEX formulas (strictly identical to the Webots repo source ones)
- [x] Create a showdown extension to convert the `$$mathContent$$` to HTML tags supported by MathJax.
- [x] Use MathJax to display the result
- [x] Put the MathJax dependency on the Cyberbotics FTP
- [x] Update the local importer script accordingly

capture d ecran 2016-04-14 a 17 48 13


fabienrohrer:
I have issues in introducing cleanly the MathJax dependency. By default, the JS is not minimized, and the dependency is huged (180Mo, >33k files). I'm trying to reduce ti to use only the required files.

fabienrohrer:
Ok, I finally reduced the dependency to the following files only:

```
+    'mathjax/2.6.1/extensions/MathEvents.js',
 +    'mathjax/2.6.1/extensions/MathMenu.js',
 +    'mathjax/2.6.1/extensions/MathZoom.js',
 +    'mathjax/2.6.1/jax/element/mml/jax.js',
 +    'mathjax/2.6.1/jax/input/TeX/config.js',
 +    'mathjax/2.6.1/jax/input/TeX/jax.js',
 +    'mathjax/2.6.1/jax/output/HTML-CSS/autoload/mtable.js',
 +    'mathjax/2.6.1/jax/output/HTML-CSS/config.js',
 +    'mathjax/2.6.1/jax/output/HTML-CSS/fonts/STIX/fontdata.js',
 +    'mathjax/2.6.1/jax/output/HTML-CSS/jax.js',
 +    'mathjax/2.6.1/MathJax.js',
```

fabienrohrer:
@omichel Could you create the symbolic link from `FTP/repository/www/mathjax` to `www.cyberbotics/mathjax` as for the other dependencies?

omichel:
Done.

fabienrohrer:
Thank you, it's working well

fabienrohrer:
👍

DavidMansolino:
:eyes:

DavidMansolino:
Just two small comments (even if I am not sure you can change this):
  - Vertical distance with array seems exaggerated:

screenshot from 2016-04-19 15 49 07
  - When there is several powers, the text tend to be very small (it was already the case before, but it seems even worse now):

screenshot from 2016-04-19 15 48 58

fabienrohrer:
> Vertical distance with array seems exaggerated:

I cannot reproduce this issue :-(
I tried to clean the cache, and on Chrome/Firefox.
The result looks much better on my environment:

capture d ecran 2016-04-19 a 17 12 10

It looks like if the MathJax CSS is not completely loaded.
Could you try again, and let me know if you have some issues in the JavaScript console?

---

> When there is several powers, the text tend to be very small (it was already the case before, but it seems even worse now):

This becomes indeed difficult to read.
I think it's more a latex issue than a MathJax issue: Latex has a syntax to solve this (`\Large`).

For example:

``` diff
diff --git a/reference/lightsensor.md b/reference/lightsensor.md
-(\cos \alpha)^{\log \frac{\log \frac{1}{2}}{2 \cdot \log ( \cos (beamWidth))}} & \scriptsize \textrm{if (SpotLight and } \alpha \leq \textrm{CutOffAngle)} \\
+(\cos \alpha)^{\Large \log \frac{ \log \frac{1}{2}}{2 \cdot \log ( \cos (beamWidth))}} & \scriptsize \textrm{if (SpotLight and } \alpha \leq \textrm{CutOffAngle)} \\
 \end{array}
```

is producing this result:
capture d ecran 2016-04-19 a 17 20 42

An alternative is to increase the font size:

```
 diff
diff --git a/css/style.css b/css/style.css
 .math {
   margin: 1em;
+  font-size: 1.2em;
 }
```

But the result seems less good looking:

capture d ecran 2016-04-19 a 17 23 44

So I propose to solve this using the first solution.

DavidMansolino:
1. I can reproduce this issue both on Firefox and Chrome, adn it seems indeed a problem with CSS loading:
```
GET https://www.cyberbotics.com/mathjax/2.6.1/jax/output/HTML-CSS/fonts/TeX/fontdata.js [HTTP/1.1 404 Not Found 1ms]
```

 2. Fixing this directly in Latex seems indeed good.

fabienrohrer:
I fixed both and it's working well on Mac and Linux now.

On Windows, I have performance issues due to the time to load the fonts which seems abnormally slow (tried with IE, edge and firefox, and with python's SimpleHTTPServer and Webots documentation server)

DavidMansolino:
I am still having problems, it seems some files are missing on the ftp:

```
downloadable font: download failed (font-family: "MathJax_Size4" style:normal weight:normal stretch:normal src index:0): bad URI or cross-site access not allowed source: https://www.cyberbotics.com/mathjax/2.6.1/fonts/HTML-CSS/TeX/woff/MathJax_Size4-Regular.woff?rev=2.6.1
```

The folder `https://www.cyberbotics.com/mathjax/2.6.1/jax/output/HTML-CSS/TeX` does not exist.

fabienrohrer:
Mmm it seems to be due to a cross-site access issue on this file:
https://www.cyberbotics.com/mathjax/2.6.1/fonts/HTML-CSS/TeX/woff/MathJax_Size4-Regular.woff

How do you test the doc?
The good way are either to use the Webots documentation server, or to launch the Python SimpleHTTPServer: https://github.com/omichel/webots-doc#run-the-doc-offline

fabienrohrer:
Anyway, I think that I didn't understand well yet the MathJax fonts. It is coming with a lot of different ways to generate the output. One of them is to have one image file per symbol per font, implying that the font directory has 30k files. This seems to be done to support old browsers..

omichel:
We should bother to ensure compatibility with old browsers, e.g. IE8, etc.

DavidMansolino:
I am launching `python -m SimpleHTTPServer` and then opening http://localhost:8000/?url=&book=reference

PR #10, opened 2016-04-15
Improve Track and TrackWheels documentation

stefaniapedrazzi:
Ditto.


stefaniapedrazzi:
The github preview works really good when editing and it is very useful because it clearly shoes which parts have been modified, added or deleted.
Probably you already know, but the same is also available when reviewing the changes by clicking on the "Display the rich diff" button.

:+1:

omichel:
:eyes: (I didn't know about the rich diff, very nice feature indeed...)

omichel:
:+1:

DavidMansolino:
> Probably you already know, but the same is also available when reviewing the changes by clicking on the "Display the rich diff" button.

Just for the note, with atom it is possible to see in real time the result of the md file with the shortcut `ctrl+shift+m`:

screenshot from 2016-04-15 15 13 30

screenshot from 2016-04-15 15 14 58

fabienrohrer:
Thank you both for the atom plugin, and the "Display the rich diff", they are both very useful.

Just a small remark about this PR: I tried to respect the style rule about having lines with no more than 80 characters. In my point of view, this is improving the readability of the MD files. Are you ok with going on with this rule? => we can discuss of this in the meeting later..

DavidMansolino:
>I tried to respect the style rule about having lines with no more than 80 characters

Yes, I think this is very good, as when I was editing the doc for the new Joystick API, I found very difficult to edit these files which has very long lines (but I am not sure we can cut those lines as they are tables):
https://github.com/omichel/webots-doc/blob/gh-pages/reference/ros-api.md

stefaniapedrazzi:
I saw no difference when writing the text or reviewing the modifications having only 80 characters or more per line.
The disadvantage of keeping the 80 characters is that every time you modify the text you will introduce additional unrelated changes just to keep the text "aligned".
So if there is no advantage in reviewing or modifying it if it less tha 8' characters I think it is simpler to remove this constraint.

PR #16, opened 2016-04-20
Improvements for an integration on the Webots website

fabienrohrer:
- [x] Fix the CSS in order to let match the MD view height with the page height => the scroll bar of the view is now matching with the main browser scrollbar.
- [x] Integrate iframe-resizer (and create a working example of iframe integration)

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -18,9 +18,8 @@ html, body {
 #center {
   left: 20%;
   width: 79%;
-  position: fixed;
-  height: 100%;
-  overflow: auto;
+  float: left;
+  position: absolute;
 }
 #navigation, #accordion, #view {
   width: 100%;
--- a/index.html
+++ b/index.html
@@ -35,7 +35,7 @@
     <script src="https://www.cyberbotics.com/jquery-ui/1.11.4/jquery-ui.min.js"></script>
     <script src="https://www.cyberbotics.com/highlight/9.2.0/highlight.min.js"></script>
     <script src="https://www.cyberbotics.com/showdown/1.3.0/showdown.min.js"></script>
-    <script src="https://www.cyberbotics.com/doc/iframeResizer.contentWindow.min.js"></script>
+    <script src="https://www.cyberbotics.com/iframe-resizer/3.5.3/iframeResizer.contentWindow.min.js"></script>
     <script src="scripts/showdown-extensions.js"></script>
     <script src="scripts/viewer.js"></script>
   </body>
--- a/index.html
+++ b/index.html
@@ -27,6 +27,7 @@
         <h2 id="title-content">Documentation</h2>
       </div>
       <div id="view"></div>
+      <div data-iframe-height></div>
     </div>
 
     <script src="https://www.cyberbotics.com/wwi/8.3/request_methods.js"></script>
@@ -34,6 +35,7 @@
     <script src="https://www.cyberbotics.com/jquery-ui/1.11.4/jquery-ui.min.js"></script>
     <script src="https://www.cyberbotics.com/highlight/9.2.0/highlight.min.js"></script>
     <script src="https://www.cyberbotics.com/showdown/1.3.0/showdown.min.js"></script>
+    <script src="https://www.cyberbotics.com/doc/iframeResizer.contentWindow.min.js"></script>
     <script src="scripts/showdown-extensions.js"></script>
     <script src="scripts/viewer.js"></script>
   </body>
--- a/local_exporter.py
+++ b/local_exporter.py
@@ -9,6 +9,7 @@ import urllib2
 dependencies = [
     'highlight/9.2.0/default.min.css',
     'highlight/9.2.0/highlight.min.js',
+    'iframe-resizer/3.5.3/iframeResizer.contentWindow.min.js',
     'jquery/1.11.3/jquery.min.js',
     'jquery/1.11.3/jquery.min.map',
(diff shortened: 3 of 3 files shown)


fabienrohrer:
I tested iframe-resizer and get quite good results with this kind of integration:

``` html
<html>
<body>
  <script src="https://www.cyberbotics.com/doc/iframeResizer.min.js"></script>
  <style>iframe{width:100%}</style>
  <iframe src="index.html"></iframe>
  <script>
    iFrameResize({
      log: true,
      heightCalculationMethod: 'max'
    });
  </script>
</body>
</html>
```

However I have blinking issues due to the animated transitions (the accordion menu, and the anchors links), and I didn't solve them yet after trying several solutions to avoid this.

fabienrohrer:
I finally found a clean solution using the "taggedElement" heightCalculationMethod and a dummy "data-iframe-height" element.
https://github.com/davidjbradshaw/iframe-resizer#heightcalculationmethod

omichel:
Can you upload a test web page on www.cyberbotics.com (FTP repository / beta) once it works?

fabienrohrer:
I have issues to put an online example until this PR is merged, because `raw.githubusercontent.com` is dening the X-Frame-Option (while `github.io` doesn't).

I propose to finalize the integration of `iframeResizer`, test locally using the following HTML file, and to merge this branch.

``` html
<html>
  <head>
    <meta charset="UTF-8">
    <style>iframe{width:100%}</style>
  </head>
  <body>
    
    <h1>Header sample</h1>
    <iframe src="index.html"></iframe>
    <h1>Footer sample</h1>

    <script src="https://www.cyberbotics.com/doc/iframeResizer.min.js"></script>
    <script>
      iFrameResize({
        checkOrigin: false,
        heightCalculationMethod: 'taggedElement'
      });
    </script>
  </body>
</html>
```

fabienrohrer:
@omichel Could you create the link from `FTP/repository/www/iframe-resizer` to `cyberbotics.com/iframe-resizer`? Thank you in advance

omichel:
Done.

fabienrohrer:
👍

fabienrohrer:
👍

omichel:
:+1:

fabienrohrer:
Ok, here is the demo:
http://www.cyberbotics.com/files/repository/files/iframe-test.html

omichel:
It looks good. I integrated it at: https://www.cyberbotics.com/doc/index.php?book=guide
Can you test it and let me know if you spot any problem? So far, it looks very good to me...
The only problem I see is on Firefox/Window, a horizontal scroll bar appears below the left hand side menu:
untitled

omichel:
It is also visible in Chrome and IE (on Windows).

omichel:
Another issue is the URL which remains the same as we navigate in the book...

fabienrohrer:
> The only problem I see is on Firefox/Window, a horizontal scroll bar appears below the left hand side menu

Ok, I will fix this.


> Another issue is the URL which remains the same as we navigate in the book...

I was expecting this kind of issues..
Moreover, the page title (which can be seen in the browser history, and in the browser tab) is also not updated.
Probably that the iframe cannot modify this. (If this is the case, a solution is to use the iframe-resizer message system to communicate this kind of messages to the container)

=> I will fix these 2 issues.

stefaniapedrazzi:
On Linux (Firefox and Chrome) I also have the horizontal scrollbar for the left hand side menu, but it doesn't seem to work for the next/previous/up links because they are one above the other if the window is small

docscreenshot

fabienrohrer:
@stefaniapedrazzi Thank you, I will try to solve this at the same time.

omichel:
I begin to believe that iframe may not be the best solution for several reasons:

1. It requires the use of iFrameResizer JS (client and server side) which is pretty heavy.
2. It doesn't solve the problem of URL and would require some complexification of iFrameResizer which to be very tricky.
3. In any case, we would need to implement some apache rewrite & php trick to get correct URL which involves the use of php files from which we need to retrieve the URL to display the correct content...

So, I will study a possible solution based on php and compare it with the iframe solution...

fabienrohrer:
I agree, I'm not very satisfied by the result too.

One important aspect to keep in mind is to facilitate the integration of the new version as much as possible, which was one big advantage of the iframe solution because we had nothing to do.

omichel:
I implemented a nice solution to preserve the URL with apache `mod_alias`:
```
AliasMatch "^/doc/(.*)$" "/var/www/doc.php"
```
Tests:
https://www.cyberbotics.com/doc/guide/webots-devices
https://www.cyberbotics.com/doc/reference/camera

doc.php:
```php
<?php
  include 'header.php';
  
  $md = substr(htmlspecialchars($_SERVER['REQUEST_URI']), 5) . ".md";
  echo "<div class='maincontent'>\n";
  echo "MD file is $md<br>\n";
  echo "</div>\n";
  
  include 'footer.php';
?>
```

omichel:
Now to be able to perform cross-domain AJAX, we should the JSONP approach which seem much simpler than the iframe / iFrameResizer approach. I am investigating this...

omichel:
Finally, it seems possible to perform cross-domain AJAX by simply modifying the HTML header with `Access-Control-Allow-Origin: *`: http://stackoverflow.com/questions/3506208/jquery-ajax-cross-domain
Strangely, it seems to work even without modifying this header...

PR #25, opened 2016-04-27
Fix wrong argument type in Supervisor documentation

stefaniapedrazzi:
This is a regression issue introduced when converting the documentation.
Webots 8.4.0 documentation is correct.

:+1:


DavidMansolino:
Have you checked if this conversion problems occurs at other places ?
If yes then :+1:

stefaniapedrazzi:
I also fixed the set MF fields.
Probably there are similar issues in other doc files but it is easier to fix them when we see the error.

:+1:

DavidMansolino:
:+1:   (I quickly checked some other functions having `[]` and they seems ok).

PR #30, opened 2016-04-27
Integration into Cyberbotics web site

omichel:
This PR fixes all the problems related with the integration of the doc on Cyberbotics web site.
Currently everything is rendered nicely on the web site. We can even view other branches very easily (see https://github.com/omichel/webots-doc/blob/gh-pages/README.md).
I merged the `style.css` into `main.css` which is the main css file of Cyberbotics' web site and committed that file.
However, I didn't yet tested if the local rendering was affected by these changes.


omichel:
The local doc viewer works again and the online view is already online at: https://www.cyberbotics.com/doc/guide

:+1:

fabienrohrer:
On Chrome, the rendering is not working as expected, while it's working on Firefox:

capture d ecran 2016-04-28 a 13 43 13

fabienrohrer:
Could you explain the modifications on the CSS?

fabienrohrer:
I'm not very confident with the merge of the 2 CSS, but I don't have good alternatives to propose.
Notably I fear that we will always tempt to drop some CSS rules vital for the Cyberbotics website.

fabienrohrer:
.. and the result is very encouraging :-)

omichel:
Well, it works on my Chrome (under Windows 10):
chrome
Can you try to empty your cache and try again?
Do you get any error message in the Javascript console?

omichel:
About the CSS, I hesitated to split it into two files, namely main.css would remain the main css of Cyberbotics web site and doc.css would contain the CSS used by the webots-doc only. However, this is not a good idea as the two CSS may conflict with each other anyway (e.g., overriding rules, etc.). So having everything in the same file is better to avoid such conflicts. We could improve things a little bit by defining a class that would inherit to all the documentation elements, e.g., <div id="view" class="webots-doc"> and have the CSS use this class to restrict the styles and avoid conflicts with the rest of the CSS.

fabienrohrer:
The cache was indeed responsible of my issues, it's now working as expected on Chrome too.

fabienrohrer:
The dynamic load of the page seems broken on the Cyberbotics website.

fabienrohrer:
Ok for the CSS.

js/viewer.js
```diff
@@ -292,21 +333,32 @@ function receiveMenuContent(menuContent) {
...
-
     updateSelection();
 }
 
+function updateMenuScrollbar() {
+    var e = document.documentElement;
+    var t = document.documentElement.scrollTop || document.body.scrollTop;
+    var p = e.scrollHeight - t - e.clientHeight;
+    // console.log("height = " + e.scrollHeight + " scrollTop = " + t + " clientHeight = " + e.clientHeight);
+    if (p < 244)
```
fabienrohrer: There are several hard coded integers in this function, and it seems important to document their origin.

fabienrohrer:
Minor note:
About coding style, and curly brackets on the one line conditions, I was following the convention we decided to adopt:
https://github.com/omichel/webots/wiki/Coding-style#javascript
I guess the reason of this choice is the tolerance to syntax error due to the fact that the language is interpreted.
However I admit I didn't follow the convention perfectly..

fabienrohrer:
Ok, the online dynamic load is fixed thanks to a4f0f3c.
It's still broken in the offline version.

omichel:
About coding styles, it is a pain to have different coding styles for C/C++ and for Javascript. For example, the use of curly brackets on one line conditions and the 4 spaces for tabulation which we don't use in C/C++... I would suggest to change this to have more consistent coding styles, e.g., to have a 2 spaces tabulation and no curly brackets for one line conditions in Javascript, like we do in C/C++.

fabienrohrer:
I'm fixing the `setupUrlLocal()` function

fabienrohrer:
I fixed `setupUrlLocal()` and finally rewrite `setupUrlOnline()` using RegExp too for homogeneity (and it offers a better expression power: the arguments may be inverted for example). The dynamic loading is working again locally and online.

Remains to fix the anchors links in the current page.

fabienrohrer:
Ok, everything seems to be fixed now. dynamic load and anchors are working online and offline.

fabienrohrer:
I updated the viewer.js on the FTP.

fabienrohrer:
👍

omichel:
:+1:

fabienrohrer:
👍

PR #40, opened 2016-04-29
Improve documentation of simulation and controller step

omichel:
Re-open #19 to me merged into the master branch


stefaniapedrazzi:
:+1:

omichel:
:eyes:

stefaniapedrazzi:
Related with changes in https://github.com/omichel/webots/pull/3748

omichel:
The `controller_synchronization.png` figure seems wrong as I believe a click on the step button should correspond to one basic time step and not to one controller step.

If we have several controllers running with different time step, this may be confusing...

stefaniapedrazzi:
>The controller_synchronization.png figure seems wrong as I believe a click on the step button should correspond to one basic time step and not to one controller step.

Yes, it's true. I will fix.

> If we have several controllers running with different time steps, this may be confusing...

I agree, but not having it is much more confusing because it is not clear at all how the synchronization works.

guide/the-user-interface.md
```diff
@@ -234,9 +239,11 @@ The `Show Distance Sensor rays` allows you to display, or to hide, the rays
 casted by the distance sensor devices. These rays are drawn as red lines (which
 become green beyond collision points). Their length corresponds to the maximum
 range of the device.
+If the distance sensor device is disabled or the first measurement is not available yet, the frustum will be drawn in gray.
```
omichel: It should be the **ray** instead of the **frustum** here.

guide/the-user-interface.md
```diff
@@ -234,9 +239,11 @@ The `Show Distance Sensor rays` allows you to display, or to hide, the rays
...
 
 The `Show Light Sensor rays` allows you to display, or to hide, the rays casted
 by the light sensor devices. These rays are drawn as yellow lines.
+If the light sensor device is disabled or the first measurement is not available yet, the frustum will be drawn in gray.
```
omichel: Same comment here.

reference/accelerometer.md
```diff
@@ -63,6 +63,8 @@ const double *wb_accelerometer_get_values(WbDeviceTag tag)
 
 The `wb_accelerometer_enable()` function allows the user to enable the
 acceleration measurement each `ms` milliseconds.
+The provided `ms` argument specifies the sensor's sampling period.
+Note that the first measurement will be available only after the sampling period has expired.
```
omichel: I would use **elapsed** here instead of **expired**.

omichel: (this is repeated many times)

stefaniapedrazzi:
:+1:

omichel:
:+1:

reference/rangefinder.md
```diff
@@ -125,6 +125,8 @@ int wb_range_finder_get_sampling_period(WbDeviceTag tag)
 
 `wb_range_finder_enable()` allows the user to enable a range-finder update each
 `ms` milliseconds.
+The provided `ms` argument specifies the sensor's sampling period.
```
fabienrohrer: I know you didn't introduce this, but I'm wondering if it won't be better to rename the `ms` argument to `refresh_rate` or so, for all the sensors, and to indicate that its unit is in milliseconds in the comments. => The argument name should describe its purpose, not its unit.

stefaniapedrazzi: If we want to change it, then it is better to use the same terminology everywhere, so using ``sampling_period``instead of ``refresh_rate`.

fabienrohrer: Yes, it would be better.

fabienrohrer: We could do this in another PR..

PR #47, opened 2016-05-03
Fix anchor scroll position

### LINKED ISSUE #46: Anchor links are not scrolling exactly to the correct position
Both the online and offline versions of the doc are affected by this problem, but not the same way...

omichel:
Fixes #46.

:+1:

### DIFF
--- a/js/viewer.js
+++ b/js/viewer.js
@@ -163,8 +163,13 @@ function redirectImages(node) {
 
 function applyAnchor() {
     var anchors = document.getElementsByName(setup.anchor);
-    if (anchors.length > 0)
+    if (anchors.length > 0) {
         anchors[0].scrollIntoView(true);
+        if (!local)
+          window.scrollBy(0, -46); // 46 is the height of the header of Cyberbotics web page
+        else
+          window.scrollBy(0, 180); // manual adjustment for the off-line version
+    }
 }
 
 function applyToTitleDiv() {


js/viewer.js
```diff
@@ -163,8 +163,13 @@ function redirectImages(node) {
 
 function applyAnchor() {
     var anchors = document.getElementsByName(setup.anchor);
-    if (anchors.length > 0)
+    if (anchors.length > 0) {
         anchors[0].scrollIntoView(true);
+        if (!local)
+          window.scrollBy(0, -46); // 46 is the height of the header of Cyberbotics web page
+        else
+          window.scrollBy(0, 180); // manual adjustment for the off-line version
```
fabienrohrer: This hard-coded value (180) seems strange. It seems that the issue is not understood correctly.
Note also there are several types of anchors, at least the titles, the API and the figures.

omichel: Yes, I don't understand why we need to set this 180 value. Do you have any idea?

fabienrohrer: No, I don't have any clue about this.
The element.scrollIntoView() function seems to be called correctly, and should work without any offset.
https://developer.mozilla.org/fr/docs/Web/API/Element/scrollIntoView

Hum, the note in this page indicating:

> L'élément peut ne pas être aligné complètement avec le haut ou le bas, selon la disposition des autres éléments.

On this link, they are recommending to use window.scrollTo() instead:
http://stackoverflow.com/questions/8922107/javascript-scrollintoview-middle-alignment

fabienrohrer: I think that the snippet proposed in answer 1 of http://stackoverflow.com/questions/8922107/javascript-scrollintoview-middle-alignment is worth to try.

omichel:
I just tried it and it doesn't work as expected. What it does is that it centers the top of the anchor in the middle of the page, which is not what we want: we want to place the top of the anchor at the top of the page. I could modify it to achieve this by removing the `- ( window.innerHeight / 2 )` subtraction and adding the offset I had before... So, it is not very helpful and I still don't understand where does this 180 pixel offset come from...

fabienrohrer:
Is the same 180 offset present in every anchor kind (figure vs title vs api)?

omichel:
Yes, it works very well for every kind of anchor...

fabienrohrer:
Could 180 be the title height? (but in this case it's strange that there is not the same issue in the online version)

omichel:
No, the title is 39 pixel height.

PR #50, opened 2016-05-04
Fix anchor scroll position (re-open)

omichel:
Re-open #47 which was closed by mistake.

### DIFF
--- a/js/viewer.js
+++ b/js/viewer.js
@@ -163,8 +163,13 @@ function redirectImages(node) {
 
 function applyAnchor() {
     var anchors = document.getElementsByName(setup.anchor);
-    if (anchors.length > 0)
+    if (anchors.length > 0) {
         anchors[0].scrollIntoView(true);
+        if (!local)
+          window.scrollBy(0, -46); // 46 is the height of the header of Cyberbotics web page
+        else
+          window.scrollBy(0, 180); // manual adjustment for the off-line version
+    }
 }
 
 function applyToTitleDiv() {


fabienrohrer:
Reading the scrollIntoView doc (https://developer.mozilla.org/fr/docs/Web/API/Element/scrollIntoView) we can maybe accept that an arbitrary offset is setup, because the doc is suggesting this..

omichel:
I don't see anything in the doc that would suggest this...

omichel:
I tried a simple example that uses our CSS and doesn't have this 180 pixel offset problem.
[test.zip](https://github.com/omichel/webots-doc/files/248368/test.zip)

fabienrohrer:
> I don't see anything in the doc that would suggest this...

They say:

> L'élément peut ne pas être aligné complètement avec le haut ou le bas, selon la disposition des autres éléments.

Meaning for me that depending on the other elements, an arbitrary offset is to be expected :-/

omichel:
For me that explains that is the page is too short, the anchor is not aligned with the top of the page. If you take my simple example, it works fine without this 180 pixel offset. I will try to investigate what are the differences between my test example and our dynamically generated pages...

omichel:
I really don't understand where does this offset come from... @fabienrohrer, when you originally developed the `viewer.js`, I guess the anchors were working with the exact vertical position without any vertical offset, right?

fabienrohrer:
Yes, I think so, but I don't know when we introduced this.

omichel:
Ok, I am giving up on this, given that it affects only the local version of the doc which is going to be discontinued soon (when we will have the embedded doc viewer in Webots).

:+1:

fabienrohrer:
👍  (no problem)

PR #60, opened 2016-05-13
Fixed code background forum

omichel:
Addresses https://github.com/omichel/webots/issues/3852
The rose background is now also used for the webots documentation, see: https://www.cyberbotics.com/doc/guide/controller-programming
This is more consistent with the forum.

:+1:

### DIFF
--- a/css/main.css
+++ b/css/main.css
@@ -716,7 +716,7 @@ a:active, a:focus {
 }
 pre {
   margin: 30px;
-  background-color: #eee;
+  background-color: #ffeeee;
 }
 code {
   background: transparent;


stefaniapedrazzi:
Maybe we could also adjust the margin. In the forum the margin seems much bigger than before merging the css with the documentation one.
On the other hand in the documentation there is no margin

doc

omichel:
As I said to @fabienrohrer, the clean long term solution is to use the HTML `class` attribute to handle this properly in the CSS file (and avoid such conflicts). I prefer to let @fabienrohrer continue with this issue as he originally designed the layout of the documentation.

omichel:
I will open an issue regarding the use of the HTML `class` attribute to improve the CSS. Meanwhile, I believe we should merge this PR as is.

:+1:

fabienrohrer:
👍

PR #90, opened 2016-06-22
Split CSS

### LINKED ISSUE #64: Use the class HTML attribute to better style the documentation
Using the `class` HTML attribute in the documentation will help us avoiding conflicts with the CSS of Cyberbotics' web site. The name of the class could be `webots-doc`. We should split the current `main.css` file into two files: `main.css` would remain the main CSS file of Cyberbotics' web site whereas `webots-doc.css` would contain only the `class` specific CSS for the Webots documentation. Only the `webots-doc.css` stylesheet file should be distributed with Webots.

fabienrohrer:
Attempt to fix #64


fabienrohrer:
👍

css/old.css
```diff
@@ -0,0 +1,210 @@
...
+.api table {
+  font-family: monospace;
+  width: 80%;
+}
+@font-face {
+	font-family: "Raleway";
+	font-style: normal;
+	font-weight: 100;
+	src:local("Raleway Thin"),local("Raleway-Thin"),url("Raleway-Thin.woff") format("woff")
+}
```
omichel: Shouldn't we simply remove this file?

css/webots-doc.css
```diff
@@ -0,0 +1,221 @@
+.webots-doc #left {
```
omichel: A header (comment) explaining what this file is about would be welcome.

index.html
```diff
@@ -6,29 +6,32 @@
...
-          <tr>
-            <td colspan="3"><a id="toc" title="Table of Contents">&equiv;</a></td>
-          </tr>
-          <tr>
-            <td><a id="previous" title="Previous page">&#x25C0;</a></td>
-            <td><a id="up" title="Up page">&#x25B2;</a></td>
-            <td><a id="next" title="Next page">&#x25B6;</a></td>
-          </tr>
-        </table>
+    <div class="webots-doc">
```
omichel: Why adding a extra level of `div` here? That seems overkill...
Can't we use simply:
```html
<div class="webots-doc" id="left">
```

css/old.css
fabienrohrer: I commited this file by mistake.

css/webots-doc.css
fabienrohrer: Ok, I just improved this.

index.html
fabienrohrer: You're right, this is not strictly necessary, but I think this is slightly cleaner and more scalable because all the doc is encapsulated in a single element: when embedding the doc in a page, it's probably easier to work on a single element (for example to set its dimension, position, etc.).

fabienrohrer:
👍

omichel:
:+1: (I am not convinced by the extra div encapsulating the doc, because it is not needed in our case and I consider it as an extra overhead ; if some user need it for layout purposes, it is very easy to add it. But OK, it's a detail).

omichel:
Done.

fabienrohrer:
perfect, thank you.

omichel:
Well, it doesn't work as expect... I am investigating why...

fabienrohrer:
What is the issue?

omichel:
The CSS was wrong. I fixed just it on the web site.
Instead of `#webots-doc` (CSS id selector), you should have used `.webots-doc` (CSS class selector).
I believe this PR should be fixed accordingly.

fabienrohrer:
I use `.webots-doc`: https://github.com/omichel/webots-doc/blob/master/css/webots-doc.css

Is your branch up-to-date?

fabienrohrer:
(I used `#webots-doc` in a previous/obsolete commit.)

omichel:
Ok, that was the problem. I uploaded the correct version now. Thank you.

omichel:
So, now the web site is using the latest version: https://www.cyberbotics.com/doc/guide
Can you check it?

PR #92, opened 2016-06-24
Renamed example.md to proto-example.md

omichel:
Better naming for the PROTO example documentation.

:+1:


DavidMansolino:
I think it is a good practice to open a corresponding PR in omichel/webots. It make sure we do not forget to update the sub-repository reference and it also allows us to launch a 'test me'.

omichel:
:+1:

DavidMansolino:
:+1:

PR #94, opened 2016-06-24
Glossary

omichel:
Addresses https://github.com/omichel/webots/issues/4004

### DIFF
--- a/reference/glossary.md
+++ b/reference/glossary.md
@@ -20,6 +20,8 @@ This glossary defines the terminology used to describe the various concepts rela
 
 **Dynamic solid**: A *dynamic solid* is a *solid* which has a [Physics](physics.md) *node* defined and can move according to the Webots dynamic physics engine (ODE).
 
+**Field**: a *field* is an attribute of a *node*. It has a name and a data type as defined in VRML97. The data type can be a single item or a list of items from *node*, boolean, integer, floating point, string, 2D vector, 3D vector, color, etc.
+
 **Kinematic solid**: A *kinematic solid* is a *solid* which has no [Physics](physics.md) *node* defined, but which can move using the Webots kinematic physics engine.
 
 **Node**: A *node* is an object used to define a simulation *world*. All the Webots *nodes* are depicted in the [Node Chart](node-chart.md) and described in details in the [Nodes and API functions](nodes-and-api-functions.md) section of this manual.
--- a/reference/glossary.md
+++ b/reference/glossary.md
@@ -2,7 +2,7 @@
 
 This glossary defines the terminology used to describe the various concepts related to a Webots simulation.
 
-**Actuator**: An *actuator* is a *node* representing a real robotics activator such as a motor, a LED, a radio emitter, etc. *Actuators* may receive commands from a *controller*. Some *actuators* are also *sensors* if they make sensory measurement, such as a motor than can provide position or force feedback.
+**Actuator**: An *actuator* is a *node* representing a real robotics actuator such as a motor, a LED, a radio emitter, etc. *Actuators* may receive commands from a *controller*. Some *actuators* are also *sensors* if they make sensory measurement, such as a motor than can provide position or force feedback.
 
 **Basic Time Step***: The *basic time step* is the time step increment used by Webots to advance the simulation time and perform physics simulation. It is specified as a field of the [WorldInfo](#worldinfo) *node* and is expressed in milliseconds.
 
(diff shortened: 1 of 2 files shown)


omichel:
:+1: (feel free to add more definitions or correct or extend the existing ones)

omichel:
:+1:

stefaniapedrazzi:
I will also add the definition of ``ancestor``, ``child`` and ``parent`` node.

stefaniapedrazzi:
A also changed *descendant* to *descendant node* given that the description explicitly refers to a node and not to the generic descendant terms (like for *kinemwtic solid*).

stefaniapedrazzi:
:+1:

omichel:
I don't fully agree on adding the *node* suffix to *descendant*, *ancestor*, *child* and *parent*. Because unlike *kinematic* or *dynamic*, these are nouns and not adjectives and it is lighter to say "the descendant of..." or the "parent of..", etc. than "the parent node of..." especially because these nouns apply on to nodes in Webots (and not to field or other concepts). For example our methods usually don't use the *node* suffix and that's right, like getParent() instead of getParentNode(), etc.

stefaniapedrazzi:
Ok, but then we should specify in the description that the description we provide is only applicable in case we refer to a node. I think it is wrong to given a generic definition of ancestor, descendant, parent and child and assume we are speaking about nodes. It could also apply to fields...

stefaniapedrazzi:
I mean the definition of "descendant" in the glossary it is not correct for me. Because we provide the specific definition of "descendant node" but it is wrong in case of "descendant field".

omichel:
Do we really need the concept of "descendant field"? That seems a bit overkill to me...

reference/glossary.md
```diff
@@ -4,13 +4,17 @@ This glossary defines the terminology used to describe the various concepts rela
 
 **Actuator**: An *actuator* is a *node* representing a real robotics actuator such as a motor, a LED, a radio emitter, etc. *Actuators* may receive commands from a *controller*. Their behavior will affect the outcome of a simulation. Some *actuators* are also *sensors* if they make sensory measurement, such as a motor than can provide position or force feedback.
 
+**Ancestor node**: An *ancestor node* is a *node* that hierarchically includes another *node* in the *scene tree*.
+
```
omichel: I would rephrase it relatively to the descendant node, e.g.:

**Ancestor**: The *ancestor* of a *descendant* is a *node* which hierarchically contains the *descendant* in the *scene tree*.

stefaniapedrazzi: It sound wrong to use ancestor and descendant without the word node. We never do it in the documentation.

If you want to define the ancestor node based on the descendant node definition is ok for me.
But this definition doesn't seems very consistent with descendant node description.
Then I would prefer something like:
*The converse notion of a descendant node.*

omichel: Yes, I agree that we should also rephrase the definition of the descendant node, so that it is consistent.

stefaniapedrazzi:
We don't need the concept of descendant field, but the word descendant in a tree could apply both to fields and nodes. The same for ancestor, child and parent and we should be consistent without the definitions.
In the scene tree the 'translation' field is a child of a 'Transform' node.
So for me a 'child' (or 'descendant') cannot be assumes to be a node.

omichel:
I would rather say that 'translation' is a field of the 'Transform' node, not a child...

omichel:
Let me speak with David to get his point of view and I will call you on Skype to finalize this discussion.

stefaniapedrazzi:
> I would rather say that 'translation' is a field of the 'Transform' node, not a child...

If you look the the hierarchy as a pure tree, then the field item is a child.
You would say that it is a field, but what will the user say at first sight?

omichel:
@stefaniapedrazzi: may I let you review my changes and let me know if you agree?

omichel:
:+1:

stefaniapedrazzi:
For me it is fine, even if I miss the reference to the 'field' intermediate structure when describing the *child* and *parent* terms.

:+1:

omichel:
I added a definition for *field*.

:+1:

stefaniapedrazzi:
:+1:

fabienrohrer:
👍

reference/glossary.md
```diff
@@ -0,0 +1,47 @@
+# Glossary
+
+This glossary defines the terminology used to describe the various concepts related to a Webots simulation.
+
+**Actuator**: An *actuator* is a *node* representing a real robotics actuator such as a motor, a LED, a radio emitter, etc. *Actuators* may receive commands from a *controller*. Their behavior will affect the outcome of a simulation. Some *actuators* are also *sensors* if they make sensory measurement, such as a motor than can provide position or force feedback.
```
DavidMansolino: > Some *actuators* are also *sensors* if they make sensory measurement, such as a motor than can provide position or force feedback.

I have a doubt about this sentence, because motor does not provide position measurement anymore, and for me the force-feedback is kind of an exception.

omichel: You are right. I will remove the position feedback from this sentence. But although it is an exception, I believe we should keep this comment about force feedback.

PR #99, opened 2016-06-27
Improved glossary definitions

omichel:
I reviewed and improved the definitions of the glossary according to comments in #94.

:+1:


DavidMansolino:
:+1:

Issue #76, opened 2016-06-10
Improve the layout


omichel:
Then if the left menu area is small the section titles are not included in the menu gray area (see BallJointParameter section title in the following screenshot):
![6c52d12c-2e23-11e6-8636-de9b681f3599](https://cloud.githubusercontent.com/assets/1264964/15970296/36df1f52-2f34-11e6-8156-e9792d147cd5.png)
Additionally if the documentation widget width is small the up, left and right arrow icons overlap.

The position of the horizontal scrollbar is a little bit weird if the page content is short.
I would probably expect it to be in any case at the bottom of the widget instead of at the end of the page content:
![dd59bff8-2e27-11e6-933c-c9813f9c1940](https://cloud.githubusercontent.com/assets/1264964/15970309/4a81b4ac-2f34-11e6-807d-52e01b43c806.png)

stefaniapedrazzi:
Same issue with the text width not matching the background box in the pink area.

stefaniapedrazzi:
Other issue:

If the documentation widget width is small the navigation arrows in the left menu overlaps.
![doc-layout-issue](https://cloud.githubusercontent.com/assets/5910449/15998950/89d6f27e-3142-11e6-8a3e-564bef4066c4.png)

stefaniapedrazzi:
To make the documentation widget usable given the small space we have now using a dock widget, we should make the left menu width manually resizeable and add the possibility to completely hide it.

omichel:
Fixed in #102.

PR #102, opened 2016-06-28
Improve layout for small resolutions

fabienrohrer:
- [x] Fix not highlighted code overflow: allow horizontal scroll when required
- [x] Fix texts of the menu overflow: crop it (horizontal scroll is not a good idea there, because scrolling vertically in the menu is recommended.).
- [x] Fix menu arrows overlap
- [x] Fix horizontal scroll bar of the menu shown for no reason
- [x] Test
    - [x] Embedded in Webots
    - [x] Firefox
    - [x] Chrome
    - [x] Safari
    - [x] IE
    - [x] Edge

### DIFF
--- a/css/webots-doc.css
+++ b/css/webots-doc.css
@@ -8,6 +8,7 @@
   width: 20%;
   position: fixed;
   height: 100%;
+  overflow-x: hidden;
   overflow-y: auto;
 }
 .webots-doc #center {
@@ -48,6 +49,7 @@
   margin: 1em;
   padding-bottom: 0;
   margin-bottom: 0;
+  overflow-x: hidden;
 }
 .webots-doc #menu ul {
   list-style-type: none;
--- a/css/webots-doc.css
+++ b/css/webots-doc.css
@@ -66,6 +66,7 @@
 .webots-doc #accordion li a,
 .webots-doc #accordion li ul li a {
   padding: 0.5em;
+  overflow-x: hidden;
 }
 .webots-doc #accordion li a {
   display: block;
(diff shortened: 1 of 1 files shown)


fabienrohrer:
The situation is much better now, but still not perfect: if the menu is too small, it becomes unreadable.

capture d ecran 2016-06-28 a 11 53 16
capture d ecran 2016-06-28 a 11 56 15
capture d ecran 2016-06-28 a 11 57 34
capture d ecran 2016-06-28 a 13 58 33


I propose to accept this PR as-is, and to explore the solution to put the menu into a resizable panel (as proposed by Stefania in #76) in another PR.

fabienrohrer:
👍

stefaniapedrazzi:
It seems that you removed the horizontal scrollbar of the left menu.
Wasn't possible to fix the issues without removing it?

Personally I would keep the horizontal scrollbar and additional make the left menu resizable.

stefaniapedrazzi:
In any case all the other issues are fixed and I confirm that now it is already better.

fabienrohrer:
A better patch is indeed probably possible.
I would propose to accept this as-is if you agree in order to not waste too much time with this, because the solution to have a resizable menu will probably completely change this part of the code.

stefaniapedrazzi:
It is ok for me to accept this PR as-it-is if the resizable menu will be introduced before releasing Webots 8.5.0.
Otherwise we should at least reintroduce the horizontal scrollbar.

:+1:

fabienrohrer:
Ok, I'm working on this this afternoon.

PR #114, opened 2016-07-05
Fix anchors and scrolling issues

### LINKED ISSUE #111: hashtags references are ignored on the first load
This URL doesn't scroll down to the `wb_robot_get_device` section, as it should:
https://www.cyberbotics.com/doc/reference/robot?version=fix-robot-doc#wb_robot_get_device

### LINKED ISSUE #112: Vertical scrollbar should be reset to the top when changing page
When you visit https://www.cyberbotics.com/doc/reference/robot and scroll down and then you click on the Camera section, the camera documentation is displayed, but the vertical scrollbar is not reset to the top as the user would expect. That looks wrong.

fabienrohrer:
- [x] Fix #112 
- [x] Remove the weird and problematic local scrolling offset
- [x] Fix #111 
- [x] Upload `viewer.js` to FTP

### DIFF
--- a/js/viewer.js
+++ b/js/viewer.js
@@ -233,6 +233,8 @@ function applyToPageTitle(mdContent) {
 }
 
 function populateViewDiv(mdContent) {
+    setupUrl(document.location.href);
+
     var view = document.getElementById("view");
     while (view.firstChild)
         view.removeChild(view.firstChild);
--- a/js/viewer.js
+++ b/js/viewer.js
@@ -543,6 +543,7 @@ function showAccodionItem(item) {
 }
 
 function getMDFile() {
+    window.scrollTo(0, 0);
     var target = computeTargetPath() + setup.page + '.md';
     console.log("Get MD file: " + target);
     $.ajax({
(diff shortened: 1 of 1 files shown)


js/viewer.js
```diff
@@ -183,8 +183,6 @@ function applyAnchor() {
         anchors[0].scrollIntoView(true);
         if (!local)
           window.scrollBy(0, -46); // 46 is the height of the header of Cyberbotics web page
-        else
-          window.scrollBy(0, 180); // manual adjustment for the off-line version
```
fabienrohrer: I think I understood the issue about this.

It's related to images, if the server is too slow to provide them, the scrolling is taking into account only the text, so the anchors above can be delayed (I can reproduce this systematically on my VM-Windows, and sometimes after clearing the cache on my Mac).
The best solution seems to scroll after the images are actually loaded.

fabienrohrer:
👍

omichel:
:eyes:

omichel:
:+1:

PR #115, opened 2016-07-05
Use jQuery to create a resizable/closable menu

fabienrohrer:
Alternative to #105

### DIFF
--- a/css/webots-doc.css
+++ b/css/webots-doc.css
@@ -20,6 +20,12 @@
   margin-bottom: 0;
   overflow: auto;
 }
+.webots-doc #handle {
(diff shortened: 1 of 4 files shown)


fabienrohrer:
This approach seems much better than before (demo with bugs: https://www.cyberbotics.com/doc/guide/sample-webots-applications)

omichel:
It looks good indeed.

fabienrohrer:
Using a jQuery resizable was not a good idea. Mainly it was impossible to set the handle at right of the scrollbar of the menu, which imply either weird behavior or ugly renderings.

PR #116, opened 2016-07-06
Use javascript/CSS to create a resizable/closable menu

fabienrohrer:
Alternative to #115

### DIFF
--- a/css/webots-doc.css
+++ b/css/webots-doc.css
@@ -28,9 +28,6 @@
   position:absolute;
   top: 50%;
 }
-.webots-doc #handle:hover {
-  background-color: #c8c8f0;
-}
 .webots-doc #center {
   top: 0px;
   left: 20%;
--- a/css/webots-doc.css
+++ b/css/webots-doc.css
@@ -12,7 +12,7 @@
 }
 .webots-doc #handle {
   left: 20%;
-  width: 8px;
+  width: 6px;
   height: 100%;
   position: fixed;
   z-index: 1;
@@ -63,10 +63,10 @@
 .webots-doc #menu {
   clear: left;
   width: calc(100% - 2em);
+  min-width: 150px;
   margin: auto;
   padding-bottom: 0;
   margin-bottom: 0;
-  overflow-x: hidden;
 }
 .webots-doc #menu ul {
   list-style-type: none;
@@ -80,7 +80,6 @@
 .webots-doc #accordion li a,
 .webots-doc #accordion li ul li a {
   padding: 0.5em;
-  overflow-x: hidden;
 }
 .webots-doc #accordion li a {
   display: block;
--- a/css/webots-doc.css
+++ b/css/webots-doc.css
@@ -10,6 +10,21 @@
   height: 100%;
   overflow-y: auto;
 }
+.webots-doc #handle {
+  left: 20%;
+  width: 10px;
+  height: 100%;
+  position: fixed;
+  z-index: 1;
+  background-color: #ededed;
+  cursor: ew-resize;
+  -webkit-box-shadow:inset 0px 0px 0px 1px #b6b6b6;
+  -moz-box-shadow:inset 0px 0px 0px 1px #b6b6b6;
+  box-shadow:inset 0px 0px 0px 1px #b6b6b6;
+}
+.webots-doc #handle:hover {
+  background-color: #c8c8f0;
+}
 .webots-doc #center {
   top: 0px;
   left: 20%;
--- a/css/webots-doc.css
+++ b/css/webots-doc.css
@@ -22,7 +22,7 @@
   border-left: 1px solid #b6b6b6;
   border-right: 1px solid #b6b6b6;
 }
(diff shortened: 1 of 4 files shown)


fabienrohrer:
Here is the best way to do this: to put a div in-between the left and the center divs with mouse events.

Demo here (still some bugs):
https://www.cyberbotics.com/doc/guide/sample-webots-applications

omichel:
That looks indeed promising...

fabienrohrer:
It seems to work well now:
https://www.cyberbotics.com/doc/guide/

**notes:**

- the internal layout of the menu has a better alignement
- the FTP files are updated
- I used a style for the handle which is close to the Webots handles:

capture d ecran 2016-07-06 a 15 46 54

fabienrohrer:
👍

stefaniapedrazzi:
It works fine, but I personally think that the minimum width is too big for the dock widget documentation.
Moreover if the the section title text is cut it seems there is not horizontal scrollbar.

fabienrohrer:
Ok, I will try to improve this.

omichel:
I agree there is something strange happening with the minimal width which is user unfriendly.
I believe we should allow the user to drag the bar to a very small size and only after he releases the mouse button we should either keep the selected size (if above the minimum) or completely hide the menu (if below the minimum). Hiding the menu while the user is dragging is user unfriendly...

I am not sure a horizontal scroll bar for the menu is useful as the user has always the possibility to drag the vertical size bar if he wants to read the full text, which is very intuitive. A scroll bar would add some unnecessary stuff in my point of view.

omichel:
The sugar would be that when releasing the resize bar at a width below the minimum, the menu would get hidden through a quick animation (like if the user had continued to drag the size bar until the menu is not visible any more). But maybe that's not strictly necessary.

omichel:
Otherwise, it looks really very good :smile:.

omichel:
Maybe another improvement to be more user friendly. When trying to go beyond the maximum menu width, we shouldn't prevent the user from further dragging the resize bar. Only after he releases it, we should resize the menu to its maximum width. This is very similar to the management of the minimum width.

fabienrohrer:
Thank you.

Just a side note before going on: I copied the min width mechanism on the scene tree view: they currently act the same (except the threshold are not identical, and the scene tree has an horizontal scroll bar)

omichel:
I see. So, maybe forget what I explained about resize management and let's try to make it similar to what happens in Qt.

omichel:
I noticed an important difference between Qt/Webots and the web resize bar:
In Qt/Webots (tested on Windows), the pane is resized to 0 only after the mouse X coordinate goes below minimum_width / 2. Whereas in the web page, that happens when the mouse X coordinates goes below minimum_width. This is a small detail, but makes it more user friendly in Qt/Webots.

fabienrohrer:
Ok, so regarding to all the comments, I propose if you agree to do a new iteration by:
- [ ] implementing the resize pane bellow `minimum_width / 2` instead of `minimum_width`
- [ ] add some horizontal scroll bars when required
.. and to let you test again if this is better or not.

omichel:
Perfect. Thanks. :smile:

fabienrohrer:
I implemented the last propositions, and still reduced the differences with Qt layout (same menu and handle width) and behavior:

capture d ecran 2016-07-06 a 17 05 40

fabienrohrer:
👍

omichel:
One remark: the equivalent sign `≡` doesn't appear in the web version, is that correct?

omichel:
The horizontal scroll bar sometimes appear bogus. In this case for example, after some random resizes, only a small part of it is visible and I cannot see more by using the vertical scroll bar. Weird.
untitled
Anyhow, I am still not convinced that this horizontal scroll bar is very useful. It makes things more complex both for us and for the user.

fabienrohrer:
> One remark: the equivalent sign ≡ doesn't appear in the web version, is that correct?

Yes, I had issues to align it vertically, and added it mainly to match the Webots layout.

> Anyhow, I am still not convinced that this horizontal scroll bar is very useful. It makes things more complex both for us and for the user.

I'm also not convinced. A good menu is supposed to be always readable; for me it's more a matter of setting a good min width.

fabienrohrer:
I dropped the horizontal scrollbar.

👍

fabienrohrer:
New TODO points after a review with Olivier:

- [ ] The handle should remain blue during the entire drag.
- [ ] When resizing the window, the menu width should stay the same.
- [ ] The handle size should be adjusted again on the Webots handle size.

fabienrohrer:
Note: the handle with on the online version was too narrow, because of bootstrap which was overriding the `box-sizing` rule.

fabienrohrer:
Note: now, the menu width is initialized at 20% of the window location, and is no more affected by the window resize, solving some resizing issues.

fabienrohrer:
👍

omichel:
:+1: Perfect!

Issue #117, opened 2016-07-06
Improve the note and code sections


omichel:
Our note and code sections are not very well looking. In particular, the note sections have a too big font size and the code sections have a too small font size... See for example https://www.cyberbotics.com/doc/reference/camera

A nice looking example of note and code sections can be found here:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Introduction_to_Object-Oriented_JavaScript

Likewise, we could use some icons, colors or background images to improve the appearance of these sections. However, it seems this mozilla layout doesn't allow code inside notes which we do allow.

omichel:
Fixed in #129, #125 & #124.

PR #118, opened 2016-07-07
Fix blockquote font-size for the online version

fabienrohrer:
Fix partially #117

### DIFF
--- a/css/webots-doc.css
+++ b/css/webots-doc.css
@@ -83,7 +83,9 @@
   border-color: #ccc;
 }
 .webots-doc blockquote {
-  font-size: 14px; // in the online doc, bootstrap is overriding font-size
+  margin: 5px 10px;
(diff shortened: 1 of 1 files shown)


fabienrohrer:
In the online doc, `bootstrap` is overriding `font-size` of the block quotes. Overriding `font-size` again in the webots CSS is a solution to solve this.

fabienrohrer:
The code doesn't have a different font-size, but looks different because a monospaced font is used.

fabienrohrer:
I also aligned the quote left border with the paragraphs.

fabienrohrer:
I propose to accept already this (as a hot fix), and to do a second iteration about improving the quotes rendering as proposed in #117.

fabienrohrer:
👍

omichel:
:+1:

PR #124, opened 2016-07-08
Update webots-doc.css

omichel:
That increases readability of code and makes the code font look at a similar size to the normal text font (already uploaded on cyberbotics' web site for testing).

### DIFF
--- a/css/webots-doc.css
+++ b/css/webots-doc.css
@@ -138,11 +138,14 @@
   border-radius: 4px;
   background-color: #eee;
   overflow-x: auto;
+  font-size: 16px;
 }
 .webots-doc code {
   padding: 0px;
   background: transparent;
   color: #444;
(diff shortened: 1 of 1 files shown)


omichel:
:+1:

DavidMansolino:
:+1:

fabienrohrer:
maybe that increasing the vertical padding and margin of the code blocks would also help, like there:
https://developer.mozilla.org/fr/docs/Web/HTML/Element/a

Currently my feeling is that some spaces are missing, rendering the visual "congested".

omichel:
I agree. I will try to improve that.

omichel:
:+1:

omichel:
:+1:

fabienrohrer:
👍  The result is very nice!

PR #129, opened 2016-07-08
Added a note icon

omichel:
Similar to notes as seen in https://developer.mozilla.org/fr/docs/Web/HTML/Element/a

:+1:


fabienrohrer:
👍

fabienrohrer:
( @omichel I allow myself to merge this PR, because of your holidays, and this is matching the new doc state. )

PR #130, opened 2016-07-11
Minor changes to the reference manual in the PROTO chapter

remidhum:
I am a bit unsure about the "pane" part, originally though it was meant to be plane but changed to panel instead.


remidhum:
:+1:

DavidMansolino:
:+1:

Issue #128, opened 2016-07-08
Visualize branches from forked repositories


DavidMansolino:
It would be nice to be able to visualize branches from forked repositories (as users will have to fork the repository to contribute and will want to see the result before raising the thumb up on the pull request).

fabienrohrer:
We could add a new GET arguments, like `?repository=remidhum`. Probably that the project name can also vary, so supporting something like `?repository=remidhum/my-custom-webots-doc-project-name` is probably required.

DavidMansolino:
We could also support to specify the repository like this: `?version=repository:branch`
This way the user simply has to copy the `from` argument of the pull request (i.e. `remidhum:language_check` see PR #126).

fabienrohrer:
Yes, this seems to be a better idea. Here too, probably that supporting a custom project name is required, so to summarize:

```
?version=[branch|tag]
?version=repository:[branch|tag]
?version=repository/custom-name:[branch|tag]
```

DavidMansolino:
> Here too, probably that supporting a custom project name is required

I agree, by default when forking the name is the same, but the user is allowed to change it afterward.
What you propose seems good.

omichel:
Done in #133.

PR #132, opened 2016-07-15
Feature supervisor hide from camera

DavidMansolino:
Linked with omichel/webots#4064

PR #133, opened 2016-07-18
Support reference to external repository

omichel:
Addresses #128.


omichel:
The Travis test is failing due to some timeout problem with `local_exporter.py`.

fabienrohrer:
It seems that http://cyberbotics.com/jquery-ui/1.11.4/theme.css is inaccessible.

omichel:
But it is accessible. I can get it...

omichel:
Yes, the URL is wrong. It should be https://www.cyberbotics.com/jquery-ui/1.11.4/theme.css

fabienrohrer:
Strange, the first time I tried, I had issues, but now I can get it too.

fabienrohrer:
This timeout is failing:
https://github.com/omichel/webots-doc/blob/master/local_exporter.py#L42

fabienrohrer:
The same timeout failed again on another file.
Maybe the Travis network is saturated or so!? Increasing the timeout should solve the issue, but it's a dirty patch.

omichel:
Shall I try this?

fabienrohrer:
Yes, gladly. I would put a timeout of 30 seconds or so.

DavidMansolino:
I have the same timeout issue on #132

omichel:
Well, it doesn't seem to fix the problem... :unamused:

omichel:
I believe it is not necessary to support a renaming of the webots-doc repository as it is very unlikely to happen (why users would do that?). Instead, they would rather create different branches.

:+1:

omichel:
:+1:

fabienrohrer:
👀

fabienrohrer:
> why users would do that?

Although it's useless, as this is a GitHub feature, soon or later a user will do it :-)
But we can indeed not take care about this very irregular case.

fabienrohrer:
👍  (seems to work like a charm)

omichel:
:+1:

PR #134, opened 2016-07-18
Fix Travis issues

fabienrohrer:
Ditto

### DIFF
--- a/local_exporter.py
+++ b/local_exporter.py
@@ -4,7 +4,7 @@
 
 import os
 import re
-import urllib2
+import urllib
 
 dependencies = [
     'highlight/9.5.0/default.min.css',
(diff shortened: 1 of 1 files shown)


fabienrohrer:
I didn't understood well the issue yet, but it's like is sometimes, the travis server fails to reach the Cyberbotics server:

`urllib2.URLError: <urlopen error ('_ssl.c:565: The handshake operation timed out',)>`

In the last test, one test passed, and the other not, meaning that the mechanism is well written.
Maybe it's due to a temporary internet "global" issue, I don't know where the Travis server is located.

fabienrohrer:
It's worth to try tomorrow.

fabienrohrer:
👍 
Hum, not very convinced, but adding a multiple trial on the urllib urlopen and read is solving the issue. I don't have better idea for now.

omichel:
:+1: That seems to be the best approach. The problem obviously lies on Travis side, not on our side. Maybe we should report it to Travis.

PR #135, opened 2016-07-19
Fixed NVIDIA spelling and other details

### LINKED ISSUE #131: Fix spelling of NVIDIA in doc
It seems that the upper case ``NVIDIA`` version is the preferred spelling.
We should check that we always use the same upper case format in everywhere in the documentation.

omichel:
Fixes #131 and other details.

:+1:


stefaniapedrazzi:
:+1:

PR #140, opened 2016-07-21
Updated Display doc related with #4071 feature

fabienrohrer:
Ditto


fabienrohrer:
👍

reference/display.md
```diff
@@ -175,10 +175,7 @@ void wb_display_detach_camera(WbDeviceTag tag);
 
 **Description**
 
-These functions are used to attach/detach a camera to a display. If a camera is attached to a display, as soon as a new image is generated by the camera it is efficiently copied in the display. This is useful to simulate mirrors for example.
-
-> **note**:
-When a camera is attached to a display it is not possible any more to draw on this display.
+These functions are used to attach/detach a camera to a display. When a camera is attached to a display, the camera images are efficiently copied to the background of the display. This is useful to simulate for example mirrors or head-up display (HUD).
```
DavidMansolino: Just for the note, an HUD is not the central mirror of the car but is mechanism like this one: http://continental-head-up-display.com/wp-content/uploads/2014/07/Head_up_display_technology_experience.jpg

fabienrohrer: Oh yes, this is clear in my mind that mirrors and HUD are different, maybe I didn't formulate this correctly..

DavidMansolino: Maybe it is me that didn't understand because I am not sure to understand why you need to attach a camera to create a head-up display ?

fabienrohrer: Yes, this is maybe a bad example. What I would like to say is something display on the top of a camera rendering, but I don't have good names for this. Maybe it's just about to tell that this is designed for drawing something on the top of a camera node, such as debug information..

DavidMansolino: Maybe using the example of displaying the result of an image processing library to draw the recognized objects on top of the camera image ?

fabienrohrer: Ok, I will try to reformulate this.

fabienrohrer:
👍

DavidMansolino:
:+1:

PR #149, opened 2016-08-05
Fix osm importer projection

DavidMansolino:
Document new argument
Linked with omichel/webots#4108

### DIFF
--- a/automobile/scenario-creation-tutorial.md
+++ b/automobile/scenario-creation-tutorial.md
@@ -20,14 +20,16 @@ This importer uses splines to improve and smooth the path of the roads, unfortun
 > **note**:
 It is strongly recommended to not use the 3D feature of the [OpenStreetMap importer](openstreetmap-importer.md) otherwise it will not be possible to add traffic using SUMO.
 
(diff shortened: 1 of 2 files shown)


DavidMansolino:
:+1:

automobile/scenario-creation-tutorial.md
```diff
@@ -20,14 +20,16 @@ This importer uses splines to improve and smooth the path of the roads, unfortun
...
 
 ```
  * map centered with this offset: x_offset_value,y_offset_value.
  * reference coordinates: latitude_value,longitude_value.
+ * projection used: 'projection parameters'.
 ```
 
-It is important that you keep these offset and reference coordinates since you will need to reuse them to generate traffic with SUMO.
+It is important that you keep these offset and projection since you will need to reuse them to generate traffic with SUMO.
+The projection parameters should looks like: `'+proj=utm +north +zone=32 +lon_0=6.505000 +lat_0=46.511000 +x_0=0 +y_0=0 +ellps=WGS84 +units=m +no_defs'`
```
omichel: looks &rarr; look

DavidMansolino:
:+1:

fabienrohrer:
👍

PR #162, opened 2016-08-18
Added important precision for the JointParameters position field

omichel:
This made me loose some time before I re-discovered this subtlety...

:+1:


fabienrohrer:
👍

Issue #158, opened 2016-08-16
Avoid using plural form of PROTO in documentation


stefaniapedrazzi:
We should not use ``PROTOs`` in the documentation and in the ChangeLog, instead it would be better to use the single form ``PROTO`` followed by a plural name like ``files`` or ``nodes.``

PR #174, opened 2016-08-31
Do not use 'PROTOs' plural form in documentation

stefaniapedrazzi:
Fix https://github.com/omichel/webots-doc/issues/158.
Related to #4188.

### DIFF
--- a/reference/slot.md
+++ b/reference/slot.md
@@ -13,15 +13,15 @@ Slot {
 added in the `endPoint` field of a [Slot](#slot) before to be able to add any
 node in the `endPoint` field of the second [Slot](#slot). Furthermore, the
 second [Slot](#slot) can be added only if it has the same `type` as the first
-one.  The [Slot](#slot) node is particularly usefull with PROTOs, it allows the
+one. The [Slot](#slot) node is particularly useful with PROTO nodes, it allows the
 user to constrain the type of nodes that can be added in an extension field of
 the PROTO. Imagine for example that you have an armed robot in which you can
 plug different kinds of hands. In order to do so you will put the hand as an
 extension field of your robot, you will then be able to add all the different
-PROTOs of hand that you have made. But nothing prevent you to add a PROTO of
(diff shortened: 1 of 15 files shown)


stefaniapedrazzi:
:+1:

DavidMansolino:
Just a small question, what is the rule for choosing between `PROTO nodes` and `PROTO models`?

stefaniapedrazzi:
There is no rule, personally I choose the one that sounds better in the sentence context.

DavidMansolino:
Ok thank you for the explanation.

:+1:

reference/slot.md
```diff
@@ -13,15 +13,15 @@ Slot {
...
 node in the `endPoint` field of the second [Slot](#slot). Furthermore, the
 second [Slot](#slot) can be added only if it has the same `type` as the first
-one.  The [Slot](#slot) node is particularly usefull with PROTOs, it allows the
+one. The [Slot](#slot) node is particularly useful with PROTO nodes, it allows the
 user to constrain the type of nodes that can be added in an extension field of
 the PROTO. Imagine for example that you have an armed robot in which you can
 plug different kinds of hands. In order to do so you will put the hand as an
 extension field of your robot, you will then be able to add all the different
-PROTOs of hand that you have made. But nothing prevent you to add a PROTO of
+PROTO models of hand that you have made. But nothing prevents you from adding a PROTO of
```
omichel: PROTO &rarr; PROTO model

omichel: I assume that we should also use PROTO model or PROTO node at the singular form, right?

stefaniapedrazzi:
Sorry, the page didn't update I didn't see the comments.
I will open a new branch.

reference/slot.md
stefaniapedrazzi: I don't know.
I would say yes for consistency, but we are not doing it for node names ('Solid', 'LightSensor', etc.).

omichel: I don't know either... But to me, it is not very important.

PR #188, opened 2016-09-12
Enhancement darwin op

daniel-cyberbotics:
Small fixes & enhancement of Darwin-OP documentation


darwin-op/gait-manager.md
```diff
@@ -12,20 +12,20 @@ default values that are stored in a ".ini" configuration file. In the
 
 **Name**
 
-**DARwInOPGaitManager(webots::Robot *robot, const std::string &iniFilename)** - *Gait Manager constructor*
+**DARwInOPGaitManager(webots::Robot \* robot, const std::string & iniFilename)** - *Gait Manager constructor*
```
DavidMansolino: Just a note, we usually don't put space after `*` and `&`.

darwin-op/motion-manager.md
```diff
@@ -1,24 +1,24 @@
...
+The `DARwInOPMotionManager` class allows you to play
 a predefined motion stored in the "motion\_4096.bin" file. The main motions and
 their corresponding ids are listed in [appendix](motions-files.md).
 
 It is also possible to add custom motions to this file by using the `Action
 Editor` tool.
 
-> **note**:
-More informations about this tool provided by ROBOTIS is available on Robotis
+> **Note**:
```
DavidMansolino: Another small note, I agree with you it will be better with an uppercase but we use everywhere in the documentation a lowercase for the note ( we should therefore change everywhere or nowhere to keep consistency)

daniel-cyberbotics: I saw an uppercase in all notes from Guide/Tutorials

stefaniapedrazzi: But in automobile, darwin-op, and reference books it seems that we always used the lower case.
I agree with David, it doesn't matter which case we use but it has to be consistent.
I will open a new issue about this.

PR #191, opened 2016-09-14
Remove 3D view toolbar expression

stefaniapedrazzi:
In documentation we still have references to ``3D view toolbar``, instead we should use ``main toolbar``.


stefaniapedrazzi:
:+1:

omichel:
:+1:

PR #192, opened 2016-09-14
Add images for Ball,Hinge,Hinge2,Slider joints

### LINKED ISSUE #178: Joint schemas in doc
_From @cypherluc on November 26, 2013 10:25_

In the reference manual, it would be nice to have schemas such as the one in the InsertNodeDialog depicting the joints visually.

_Copied from original issue: omichel/webots#84_

daniel-cyberbotics:
Fixes #178 
:+1:


daniel-cyberbotics:
fixes #178

omichel:
When comparing the Hinge2 drawing to the one of ODE, I am wondering if they are equivalent... In the one of ODE, it is clear that it can be used to model the directional wheel of a car. With yours, it is not clear. Also, in the one of ODE, the two axis cross each other, which is not the case with yours.

omichel:
There are strange shadows on the top of the sphere. Can't we remove them?

omichel:
I am wondering if we should add some text for specifying the parent solid, solid endpoint, anchor, axis, etc.

daniel-cyberbotics:
:+1:

omichel:
I would remove the word "representation" in the figure labels.

stefaniapedrazzi:
The joint field is called ``endPoint`` with capital ``P``, because we use the two separate words form.
Is not better to write ``EndPoint`` or ``End point`` in the images to be consistent?

daniel-cyberbotics:
:+1:

PR #195, opened 2016-09-21
Protect underscore in shaderpart.md

fabienrohrer:
As reported in #193, deal better with the special MD characters by protecting them into back quotes.

Result here: https://www.cyberbotics.com/doc/reference/shaderpart?version=fix-underscores

PR #196, opened 2016-09-21
Fix table rendering in Github view.

fabienrohrer:
Fixed comment in #195: the table of the shaderpart.md file is not rendered correctly in the Github view only.

Result: https://github.com/omichel/webots-doc/blob/fix-table-for-github-rendering/reference/shaderpart.md


fabienrohrer:
👍

PR #199, opened 2016-09-22
Fix tutorial

vincentdm05:
Information complement for new Solid node shading (no shading by default) etc...


vincentdm05:
👍

fabienrohrer:
👀

guide/tutorial-2-modification-of-the-environment-20-minutes.md
```diff
@@ -115,6 +115,9 @@ Finally add a Physics node to the `physics` field of the Solid. By modifying the
 *{0, 0.1, -0.2}* for example). Save the simulation. The result is depicted in
 [this figure](#your-first-rigid-body-in-webots).
 
+> **Note**:
```
fabienrohrer: This "note" seems to be required to do the tutorial correctly, so logically it should not be a note (which is kept for doing a precision), but it should be integrated in the hands-on just above.

guide/tutorial-4-more-about-controllers-20-minutes.md
```diff
@@ -170,7 +170,7 @@ sensors as follows:
 >   "ps4", "ps5", "ps6", "ps7"
 > };
 >
-> for (i=0; i<8; i++) {
+> for (int i=0; i<8; i++) {
```
fabienrohrer: In the majority of C compilers, this will leads to a warning (or even errors?), because the `i` variable cannot be defined in the `for` statement:

```
int i; // define int above.
for (i=0; i<8; i++) {
```

.. and `int i;` is already defined above in the code.
Same comment for the other modifications bellow.

guide/tutorial-3-appearance-15-minutes.md
```diff
@@ -68,11 +68,12 @@ The aim of this subsection is to apply a texture on the ball. A texture on a
...
-Similarly add an Appearance node to the ball. Instead of a Material node, add an
-ImageTexture node to the `texture` field of the Appearance node. Add an item to
-the `url` field using the `Add` button. Then set the value of the newly added
-`url` item to "WEBOTS\_HOME/projects/default/worlds/textures/bricks.png" using
-the file selection dialog.
+Similarly add an Appearance node to the ball. Add an ImageTexture node to the 
+`texture` field as well as a Material node to the `material` field of the 
+Appearance node. Add an item to the `url` field using the `Add` button. Then set 
+the value of the newly added `url` item to 
+"WEBOTS\_HOME/projects/default/worlds/textures/bricks.png" using the file
```
fabienrohrer: Is `bricks.png` correct here? (You changed it to `red_brick_wall.png` just bellow)

guide/tutorial-2-modification-of-the-environment-20-minutes.md
stefaniapedrazzi: We could also update the screenshots if they are no longer up-to-date with the expected scene status

vincentdm05: I agree. However the screenshot is not correct in this case, it should not display a NULL appearance. Should I replace it?

fabienrohrer: I would say: the simplest solution for you to have a coherent tutorial. An alternative is to change the position of reference to the image to let this match (if possible?).

vincentdm05: You mean crop it to only see the view without the panel on the left?

stefaniapedrazzi: I would in any case add a new screenshot with the white/NULL appearance.
This is one of the first tutorials and it is better to really document step by step what you should see in the scene tre.

vincentdm05: Good idea, but for coherence it should be an ubuntu window and I don't have the means to do it easily. Is that really a problem?

stefaniapedrazzi: Personally I don't like to mix different platform screenshots.
If only the 3d scene changed and you can just replace it then it is ok to modify the existing screenshot.

vincentdm05: Unfortunately it's not only the 3d view, and the font is not the same in either Windows or Mac...

vincentdm05: I could change the tutorial slightly such that only the 3d view has to be modified and I can add another image of a new 3d view of the shaded object.

fabienrohrer: It seems good, thank you.

stefaniapedrazzi: The screenshots are old (Webots 7), I think it is worth updating all of them especially the ones showing the Webots version.

stefaniapedrazzi: ok

vincentdm05:
👍

guide/tutorial-2-modification-of-the-environment-20-minutes.md
```diff
@@ -108,15 +108,15 @@ As Geometry nodes we will use Spheres.
...
-Finally add a Physics node to the `physics` field of the Solid. By modifying the
-`translation` field of the Solid node, place the ball in front of the robot (at
-*{0, 0.1, -0.2}* for example). Save the simulation. The result is depicted in
-[this figure](#your-first-rigid-body-in-webots).
-
-> **Note**:
-In order to get a shaded object as shown in the following image, it is necessary to also add an appearance. To do this, select the `appearance` field of the Shape node and use the `Add` button. Then, you can add a blank `Material` to the `material` field in a similar way.
+Shape node to it. In order to get a shaded object as shown in the figure, it is 
+necessary to specify an appearance. To do this, select the `appearance` field of 
+the Shape node and use the `Add` button. Then, you can add a blank `Material` to
```
fabienrohrer: The term `blank Material` sounds not correct here, it sounds like "white" to me, while the default material is grey. => `default` Material would be better in my point of view.

vincentdm05:
👍

PR #209, opened 2016-09-28
Documentation for Robotis OP2

daniel-cyberbotics:


daniel-cyberbotics:
:+1:

robotis-op2/bibliography.md
```diff
@@ -0,0 +1,17 @@
+# Bibliography
+
+For any information about Webots, please visit [Cyberbotics
+website](http://www.cyberbotics.com).
+
+For any information about the Robotis-OP 2 robot, please visit [ROBOTIS
```
stefaniapedrazzi: The correct spelling is not ROBOTIS OP2?
Same comment for the other files.

daniel-cyberbotics: Right, actually we kept Robotis OP2 for better readability

robotis-op2/keyboard-inputs-are-not-working-in-remote-compilation-and-or-result-in-a-segmentation-fault.md
```diff
@@ -0,0 +1,11 @@
+## Keyboard inputs are not working in remote compilation and/or result in a segmentation fault
+
+Check that the keyboard window is open on the robot when enabling the
+keyboard and be sure this window to not lost keyboard focus on the robot.
```
stefaniapedrazzi: to not lost  -> do not lost ?

robotis-op2/menu.md
```diff
@@ -15,19 +15,21 @@
     - [Walk](walk.md)
     - [Soccer](soccer.md)
 - [Robot Window](robot-window.md)
-    - [Accelerometers](accelerometers.md)
-    - [Cameras](cameras.md)
-    - [Gyro](gyro.md)
-    - [Servos](servos.md)
-- [Cross-Compilation](cross-compilation.md)
+    - [Accelerometer](accelerometer.md)
+    - [Gyroscope](gyroscope.md)
```
stefaniapedrazzi: The title in gyroscope.md is still Gyro.
Note that Gyro is the Webots device name.

robotis-op2/menu.md
```diff
@@ -15,19 +15,21 @@
...
+- [Remote compilation](remote-compilation.md)
     - [Settings](settings.md)
     - [Send a controller to the robot](send-a-controller-to-the-robot.md)
     - [Permanently install a controller to the robot](permanently-install-a-controller-to-the-robot.md)
     - [Uninstall Webots files from the robot](uninstall-webots-files-from-the-robot.md)
     - [Dynamixel MX28 firmware](dynamixel-mx28-firmware.md)
     - [Using keyboard](using-keyboard.md)
 - [Remote control](remote-control.md)
-    - [Using remote-control from robot-window](using-remote-control-from-robot-window.md)
+    - [Using remote control from Robot Window](using-remote-control-from-robot-window.md)
```
stefaniapedrazzi: same comment as before about capital letters in robot window

robotis-op2/motion-editor.md
```diff
@@ -0,0 +1,16 @@
+## Motion editor
+
+This tab allows you to create, edit or play a motion file. It contains an animation
+which can be executed by the robot at any time.
+
+%figure "Motion editor tab of the Robot Window"
```
stefaniapedrazzi: same comment as before about capital letters in robot window

robotis-op2/motion-editor.md
```diff
@@ -0,0 +1,16 @@
...
+![Motion Editor view](images/window_motion_editor.png)
+%end
+
+The motion is composed of multiple poses (Left pane).
```
stefaniapedrazzi: why Left (and Right just after) have capital letter?

robotis-op2/motion-editor.md
```diff
@@ -0,0 +1,16 @@
...
+A pose is an instant in the time the robot has to follow.
+The robot will smoothly move from one pose to the next one.
+
+A pose is composed of multiple states (Right pane). A state represent a position of one motor.
```
stefaniapedrazzi: represent ->represents

robotis-op2/robot-window.md
```diff
@@ -0,0 +1,18 @@
+# Robot Window
+
+When you double click on the robot, a new window appears. This window is called
+`Robot Window`, it has several tabs allowing you to perform different things.
```
stefaniapedrazzi: Note that in the User Guide where the robot window is described no capital letters are used

robotis-op2/robot-window.md
```diff
@@ -0,0 +1,18 @@
...
+First tabs concern the simulation. They will be described here, the last tab is
```
stefaniapedrazzi: A '.' will be better than ',' after ``here``.

robotis-op2/robot-window.md
```diff
@@ -0,0 +1,18 @@
...
+used to interact with the real robot and will therefore be described in next
+sections.
+
+You can at any time restart the `Robot Window` by pressing the revert button of
```
stefaniapedrazzi: I don't understand what means restarting the robot window

robotis-op2/robot-window.md
```diff
@@ -0,0 +1,18 @@
...
+the simulation.
+
+## Sections
+- [Accelerometer](accelerometer.md)
+- [Gyroscope](gyroscope.md)
```
stefaniapedrazzi: Same comment as before Gyro vs. Gyroscope

robotis-op2/simulated-devices.md
```diff
@@ -37,27 +32,56 @@ robot, they are called `BackLedGreen`, `BackLedBlue` and `BackLedRed`.
...
 
-%end
+The three other LEDs are unicolor and placed on the robot back.
+They are called `BackLedGreen`, `BackLedBlue` and `BackLedRed`.
+
+### Accelerometer
+
+The accelerometer returns values between 0 and 1024 corresponding to values
+between -3 [g] to +3 [g] like on the real robot.
+Its name is `Accelerometer`.
```
stefaniapedrazzi: To me this sentence is not very clear. I would rather write something like:
``The Webots device simulating it is called 'Accelerometer'``

Same for the gyroscope and speaker

robotis-op2/simulated-devices.md
```diff
@@ -37,27 +32,56 @@ robot, they are called `BackLedGreen`, `BackLedBlue` and `BackLedRed`.
...
+### Speaker
+
+The speaker name is `Speaker` and it can be used to play sound files or perform
+text-to-speech.
+
+
+
+
+
+
```
stefaniapedrazzi: why so many empty lines?

robotis-op2/simulated-devices.md
```diff
@@ -37,27 +32,56 @@ robot, they are called `BackLedGreen`, `BackLedBlue` and `BackLedRed`.
...
+### More information
```
stefaniapedrazzi: this section still refers to the DARwInOP and not to ROBOTIS OP2

robotis-op2/bibliography.md
stefaniapedrazzi: Ok, so here it is wrong because of the '-'.

But personally I think it is better to use the correct spelling.
Note that also the book title doesn't contain the correct spelling of the robot name.

robotis-op2/vision-manager.md
```diff
@@ -0,0 +1,88 @@
...
+- The height of the image.
+- The color hue of the target object to find.
+- The tolerance on the color hue of the target object to find.
+- The minimum color saturation of the target object to find.
+- The minimum color value of the target object to find.
+- The minimum percentage of color value in the image to validate the result.
+- The maximum percentage of color value in the image to validate the result.
+
+To find the color hue of the target object and to understand the impact of the
+saturation and value you can refer to figure [HSV](#sv-colorspace) , for more
```
daniel-cyberbotics: so the previous missing space was here ^^

robotis-op2/accelerometer.md
```diff
@@ -3,19 +3,16 @@
...
+types of graph can be plotted. The first three are one axis in function of an
 other, and the last one, plots the value of the three axes in function of the
 time. The corresponding colors are the following:
 
 - Red for axis X
 - Green for axis Y
 - Blue for axis Z
 
-%figure "Accelerometer tab of the robot-window"
+%figure "Accelerometer tab of the Robot Window"
```
daniel-cyberbotics: Right, I will open another PR to make other books consistent.
I think using capital letters is good teaching method, because it emphasizes the Webots concept.

stefaniapedrazzi: Personally I prefer the lower case version

stefaniapedrazzi: I think that using capital letter is sometimes disturbing and the sentence is much more readable without them

robotis-op2/motion-editor.md
daniel-cyberbotics: to test my reviewer :)

robotis-op2/robot-window.md
daniel-cyberbotics: There were capital letters before I arrived, so I just used it and make the rest consistent to it...

robotis-op2/robot-window.md
daniel-cyberbotics: reset ?

robotis-op2/simulated-devices.md
daniel-cyberbotics: oups!

robotis-op2/bibliography.md
daniel-cyberbotics: where is the book title ?

robotis-op2/robot-window.md
stefaniapedrazzi: In this chapter, to me it would be ok to use once the version with the capital letter because it really describes what it is.
But in other chapters where the focus is on something else, using lower case make the text more readable.

robotis-op2/robot-window.md
stefaniapedrazzi: In this sentence I don't understand what is the point of restarting the robot window and why someone should need it.

In this case the previous sentence was more understandable:
``In the unlikely case of something going wrong with the Robot Window (freeze, bad behavior, etc.) you can at any time restart it by pressing the revert button of the simulation.``

robotis-op2/bibliography.md
stefaniapedrazzi: Here:
https://github.com/omichel/webots-doc/blob/9c91de9d0125bbd70b01b7cee687eda3c2b6b228/js/viewer.js#L211

robotis-op2/robot-window.md
daniel-cyberbotics: Yes. I was thinking about a user reading that and saying to himself: "This is a professional software, and it seems they know it can freeze whereas it should not badly behave..."
So I removed it but I let the information that it can be restarted (the user will refer to this point in the unlikely case something is going wrong).
What do you recommend to do ?

stefaniapedrazzi: Maybe I would simply remove the whole sentence

daniel-cyberbotics:
:+1:

stefaniapedrazzi:
@all: what do you think about using ``ROBOTIS OP2`` vs. ``Robotis OP2``?
With the DARwIn-OP and all the others robots and company names we try to be strict about the spelling.

omichel:
`ROBOTIS OP2` is indeed the official spelling for the robot. Therefore, it is good to use it. However, we could also consider that Robotis is the company name and OP2 is the robot name. And usually, we write company names not all caps, e.g., MICROSOFT logo is all-caps, but we often write Microsoft because it's more readable... So, I would maybe prefer `Robotis OP2` which is more readable. In any case, we should choose one spelling and keep consistent with it.

stefaniapedrazzi:
Ok, so following your idea if Robotis is the name of the company then we could also simply call the robot OP2. Why put the name of the company everywhere?
On the other hand if the name of the product is ROBOTIS OP2 then we should use the correct spelling.

stefaniapedrazzi:
Just note that we write NVIDIA in upper case everywhere.

omichel:
Yes, I agree, this is not very consistent... We should write `Nvidia` instead as it is done on Wikipedia: https://en.wikipedia.org/wiki/Nvidia. Using OP2 for the name of the robot is not a good idea as it is almost never referred to as such and people won't realize it corresponds to this robot.

I believe we had similar hesitations with `DARwIn-OP` and `youBot`. I believe that we should consider these spellings as logos, rather correct spelling and use a simple plain spelling, like `Darwin OP`, `Youbot`, `Robotis OP2`...

stefaniapedrazzi:
Personally I don't agree.
We can be consistent only using the correct spelling.

So the why writing ``OP`` and not ``Op``, what it is the rule? and if there is no rule then it will only be a big mess..

omichel:
Ok, so I would suggest to open a new issue on general robot and brand naming conventions in order to define a clear rule on how to proceed with this topic. Meanwhile, this shouldn't be a blocking issue for Webots 8.5.0. I am fine with the `ROBOTIS OP2` naming as well. So maybe it safer to go it until we make a general decision and fix other brand names / robot names to be consistent.

daniel-cyberbotics:
Finally :)
:+1:

PR #213, opened 2016-10-10
add composed cube map texture documentation

vincentdm05:
Documentation for the new WbComposedCubeMapTexture node.


stefaniapedrazzi:
The origin/develop branch is not up-to-date with the origin/master changes.
But after updating the origin/develop branch (see  #214), this branch should be ok.

For information: the test failed because the new page was not references from the nodes submenu.

PR #215, opened 2016-10-10
add composed cube map texture documentation

vincentdm05:
Adds documentation for the new WbComposedCubeMapTexture node.


vincentdm05:
👍

reference/composedcubemaptexture.md
```diff
@@ -0,0 +1,116 @@
...
+  vrmlField SFNode front   NULL
+  vrmlField SFNode back    NULL
+}
+```
+
+### Description
+
+The [ComposedCubeMapTexture](#composedcubemaptexture) node defines a cubic texture used for environment mapping.
+
+Each of the six fields takes a texture corresponding to the side of the environment. For instance `right` takes the texture displayed on the side of the cube along the +X direction. Accordingly we have the following order, with the Z-axis pointing outwards: right:+X (0), left:-X (1), top:+Y (2), bottom:-Y (3), front:+Z (4), back:-Z (5).
```
omichel: I would mention what types of nodes are accepted in the fields of `ComposedCubeMapTexture`. From the example, I guess it accepts only `ImageTexture` nodes, but this should be explicit in the description.

omichel: Also, what if a field is left to `NULL`? We should describe what happens in that case.

omichel: I am not sure what you mean with "with the Z-axis pointing outwards"....

vincentdm05: Is "the Z-axis being the result of the cross-poduct of Y by X" brief and clear enough?

stefaniapedrazzi:
Maybe it could be interesting to also add a screenshot of the result given by the Shape example.

reference/composedcubemaptexture.md
omichel: Yes, it is clear, but there is no need to mention this, as we always refer to a standard coordinate system where Z is the result of the cross-product of Y by X...

vincentdm05: I wrote "pointing forward (towards the screen)" in the end, is that good or is it better not to specify anything?

omichel: I would simply drop it. The user is supposed to know that objects have a local coordinate system.

vincentdm05:
👍

reference/composedcubemaptexture.md
```diff
@@ -0,0 +1,125 @@
...
+  vrmlField SFNode front   NULL
+  vrmlField SFNode back    NULL
+}
+```
+
+### Description
+
+The [ComposedCubeMapTexture](#composedcubemaptexture) node defines a cubic texture used for environment mapping. It is composed of exactly six [ImageTexture](imagetexture.md) nodes, all of which have to be defined for the resulting texture to be generated.
+
+Each of the six fields takes a texture corresponding to the side of the environment. For instance `right` takes the texture displayed on the side of the cube along the +X direction. Accordingly we have the following order: right:+X, left:-X, top:+Y, bottom:-Y, front:+Z, back:-Z.
```
omichel: I would add a space between the colon `:` and axis, e.g., "right: +X, left: -X, etc.".

reference/composedcubemaptexture.md
```diff
@@ -0,0 +1,125 @@
...
+  vrmlField SFNode top     NULL
+  vrmlField SFNode bottom  NULL
+  vrmlField SFNode front   NULL
+  vrmlField SFNode back    NULL
+}
+```
+
+### Description
+
+The [ComposedCubeMapTexture](#composedcubemaptexture) node defines a cubic texture used for environment mapping. It is composed of exactly six [ImageTexture](imagetexture.md) nodes, all of which have to be defined for the resulting texture to be generated.
```
omichel: I would add a sentence, like "If an ImageTexture node is missing, ..." to explain what happen in this case. By default all the textures are set to `NULL`, so it shouldn't crash in that case. So, what if a texture or all textures are missing. Is it using a while or black texture instead? Is the result undefined?

vincentdm05:
👍

reference/composedcubemaptexture.md
```diff
@@ -0,0 +1,126 @@
...
+  vrmlField SFNode top     NULL
+  vrmlField SFNode bottom  NULL
+  vrmlField SFNode front   NULL
+  vrmlField SFNode back    NULL
+}
+```
+
+### Description
+
+The [ComposedCubeMapTexture](#composedcubemaptexture) node defines a cubic texture used for environment mapping. It is composed of exactly six [ImageTexture](imagetexture.md) nodes. If one of the [ImageTexture](imagetexture.md)s is not set, the corresponding side of the cubic texture will be black.
```
fabienrohrer: I think it would be better to mention environment mapping later in the description as an utilisation example, but not in the first sentence description: this node deals with cubic texture, nothing more.

fabienrohrer: Telling that the six textures are matching with the cube faces (without entering in details here) here would improve the understanding of the node.

reference/composedcubemaptexture.md
```diff
@@ -0,0 +1,126 @@
...
+
+Each of the six fields takes a texture corresponding to the side of the environment. For instance `right` takes the texture displayed on the side of the cube along the +X direction. Accordingly we have the following order: right: +X, left: -X, top: +Y, bottom: -Y, front: +Z, back: -Z.
+
+The [ComposedCubeMapTexture](#composedcubemaptexture) is typically used together with a [ComposedShader](composedshader.md). It can be set to an Appearance's texture field on its own, or in conjunction with other textures inside a [MultiTexture](multitexture.md) node.
+
+In the fragment shader the cubic texture can be queried through a `uniform samplerCube` using the OpenGL function `vec4 textureCube (samplerCube sampler, vec3 coord [, float bias] )`.
```
fabienrohrer: This seems very good. Could you add a similar sentence in reference/imagetexture.md?

reference/composedcubemaptexture.md
```diff
@@ -0,0 +1,126 @@
...
+}
+```
+
+### Description
+
+The [ComposedCubeMapTexture](#composedcubemaptexture) node defines a cubic texture used for environment mapping. It is composed of exactly six [ImageTexture](imagetexture.md) nodes. If one of the [ImageTexture](imagetexture.md)s is not set, the corresponding side of the cubic texture will be black.
+
+Each of the six fields takes a texture corresponding to the side of the environment. For instance `right` takes the texture displayed on the side of the cube along the +X direction. Accordingly we have the following order: right: +X, left: -X, top: +Y, bottom: -Y, front: +Z, back: -Z.
+
+The [ComposedCubeMapTexture](#composedcubemaptexture) is typically used together with a [ComposedShader](composedshader.md). It can be set to an Appearance's texture field on its own, or in conjunction with other textures inside a [MultiTexture](multitexture.md) node.
```
fabienrohrer: I think that the doc of this node should not refer to Appearance, but Appearance should explain how we can include a ComposedCubeMapTexture within (because Appearance uses the ComposedCubeMapTexture, not the inverse). Some modifications in reference/appearance.md are probably required to explain that ComposedCubeMapTexture is supported in Appearance.texture

fabienrohrer:
The update of the nodes_hierarchy.png file (https://github.com/omichel/webots-doc/blob/master/reference/images/node_hierarchy.png) is missing here, while you modified it in the webots branch.

reference/composedcubemaptexture.md
vincentdm05: This node actually does deal with environment mapping specifically, much like the corresponding X3D node does: http://doc.x3dom.org/author/CubeMapTexturing/ComposedCubeMapTexture.html
Furthermore, the texture is mirrored such that it is convenient for reflection mapping, and not for basic texturing.
Therefore I think it is worth mentioning early.

vincentdm05: The correspondence between texture and cube face is specified in the next paragraph, which I think goes into necessary depth.

vincentdm05:
How do you produce such a node_hierarchy.png? I am trying to export the .dia with `dia` but the result is very pixelated, not at all as good as the original...

vincentdm05:
👍

PR #220, opened 2016-10-18
Moved source files from webots repository

omichel:
👍


stefaniapedrazzi:
`src/doc/dia/.gitignore` (ignoring generated *.pdf and *pdf.png) is not copied

omichel:
I added the `.ignore` files.
I will remove all these files in https://github.com/omichel/webots/pull/4478

👍

stefaniapedrazzi:
I'm not sure about the folder renaming, i.e. using the program name instead of the format.
The `svg` files can be modified using other tools than `Inkscape`.

Then, the `formula` folder name was more understandable than `tex`, given that the TEX file names doesn't specify that they contain formulas.

stefaniapedrazzi:
By the now `tex`is the only folder that uses a format name instead of a program name.

omichel:
`tex` is a program name. Although in this case, it is probably one of its friends that is used, e.g. `pdftex` or `latex`.

omichel:
I renamed the folders following your remarks.

👍

src/dia/Makefile
```diff
@@ -0,0 +1,79 @@
...
+ tutorial_4_wheels_lowlevel.pdf  \
+ tutorial_4_wheels_lowlevel.pdf.png  \
+ tutorial_ackermann_vehicle.pdf \
+ tutorial_ackermann_vehicle.pdf.png \
+ controller_plugin.pdf \
+ controller_plugin.pdf.png
+
+install : all
+	cp $(REF_TARGETS) ../../../doc/reference/pdf
+	cp $(GUIDE_TARGETS) ../../../doc/guide/pdf
```
stefaniapedrazzi: These pathe seem to refer to the old documentation

src/Makefile
```diff
@@ -0,0 +1,20 @@
...
+	@echo "  make math                  for the math formulas"
+	@echo "  make clean                 remove useless files"
+
+dia: dia/*.dia
+	@echo "Producing diagrams..."
+	@make install -C dia
+
+math: math/*.tex
+	@echo "Producing math formulas..."
+	@make install -C math
```
stefaniapedrazzi: I get an error when executing `make math` in `src` folder.
But it works fine if I execute `make` directly in the `math` folder.

omichel:
That should be all fixed.

👍

PR #224, opened 2016-10-24
Document Focus feature

daniel-cyberbotics:


daniel-cyberbotics:
:+1:

guide/the-scene-tree.md
```diff
@@ -30,6 +30,8 @@ value.
 ![](images/help-button.png =26x26) `Help`: Context sensitive help for the
 currently selected node.
 
+![](images/focus_button.png =26x26) `Focus`: Move viewpoint to focus on selected node.
```
daniel-cyberbotics: Moving the viewpoint actually moves the scene ?
Like saying the world moves and not the viewer ?

stefaniapedrazzi: It was just an example of what I would write not exactly the sentence to write.
But I think that it is important to write at the results is that the object is at the center of the displayed image. You can use other words than scene if you think that they are better.

daniel-cyberbotics: ah yes, you're right ! currently it is quite a circular definition ^^

daniel-cyberbotics:
:+1:

stefaniapedrazzi:
Please make sure that the screenshots are consistent with the real behavior before merging (disabled buttons).
Is it normal that the disable button is surrounded by a lighter grey area?

daniel-cyberbotics:
> Is it normal that the disable button is surrounded by a lighter grey area?

You have a very good screen to see that ! I will fix that.

stefaniapedrazzi:
But I'm wondering if it wouldn't be better to let the button always enabled independently if a field or node is selected and automatically search for the upper node that can be focused.

daniel-cyberbotics:
In my country we say that "the best is the enemy of the good". Can we discuss of that enhancement later ?
Also it seems to be confusing to have an item toolbar button which applies to a parent item. Maybe it would break consistency with the 'reset' and 'help' button ?

daniel-cyberbotics:
Silly me! there is no consistency between 'reset' and 'help' buttons, and they apply to parent item.
Good enhancement to integrate in this feature so :)

stefaniapedrazzi:
After modification a review is required again before merging.

daniel-cyberbotics:
Sorry, I was thinking about updating doc submodule...
Do you have any review ?

stefaniapedrazzi:
No, it's fine.

omichel:
`focus_button.png` should be renamed.

PR #229, opened 2016-11-01
Update ROS version

stefaniapedrazzi:
Related to https://github.com/omichel/webots/pull/4528.

On Ubuntu 16.04 packages the default ROS version is `kinetic` and won't test with other versions.


stefaniapedrazzi:
:+1:

PR #231, opened 2016-11-09
Mac: Update the documentation of the Webots installation

fabienrohrer:
Doc of https://github.com/omichel/webots/pull/4565

Result: https://www.cyberbotics.com/doc/guide/installation-procedure?version=fix-sierra-support


fabienrohrer:
Status: to document: the right-click + open procedure.

omichel:
In order to better document it for people who never read the docs, I propose to modify the MacOSXBackgroundImage in `src/packaging` to something like this:
macosxbackground
(the window should be also resized accordingly)
Here are my source images in case you want to improve on it (or add them in the manual):
right_click
open

fabienrohrer:
Ok, I'm doing this right now.
The instructions into the background image is a very good idea.

I already done this image which shows the operation in a single screenshot:

mac-open-anyway

daniel-cyberbotics:
It would be nice to make the rectangle green and not red. Red is like "dangerous", green is like "Ok, no problem !"

fabienrohrer:
I think the doc is ok:

👍

guide/installation-procedure.md
```diff
@@ -170,16 +170,28 @@ kern.sysv.shmseg=32
...
+During the first launch of Webots, Mac OS X may complain about the opening of Webots because it is from an unidentified developer
+(see [this figure](#unidentified-developer-dialog)).
+
+%figure "Unidentified developer dialog"
+
+![mac-unidentified-developer-dialog.png](images/mac-unidentified-developer-dialog.png)
+
+%end
+
+In this case, right-click on the Webots icon, and select the `open` menu item.
```
omichel: I would add "(or Ctrl-click)" and a capital letter to `Open`:
In this case, right-click (or Ctrl-click) on the Webots icon, and select the `Open` menu item.

guide/installation-procedure.md
```diff
@@ -170,16 +170,28 @@ kern.sysv.shmseg=32
...
+Mac OS X should propose to open the application anyway
+(see [this figure](#unidentified-developer-dialog)).
+
+%figure "Open Webots anyway"
+
+![mac-open-anyway.png](images/mac-open-anyway.png)
 
 %end
+
+In earlier versions of Mac OS X, this last operation may not work.
```
omichel: Do you mean new versions coming after Sierra?

guide/installation-procedure.md
```diff
@@ -170,16 +170,28 @@ kern.sysv.shmseg=32
...
+In this case, refer to your Mac OS X security settings to open Webots anyway
+(`Preferences / Security & Privacy / General / Allow apps downloaded from:`).
```
omichel: It is "System Preferences" instead of "Preferences".

guide/installation-procedure.md
```diff
@@ -170,16 +170,28 @@ kern.sysv.shmseg=32
...
-In this case, please modify the `Preferences / Security & Privacy` OS settings
-as shown in the figure.
-Don't forget to restore the settings once Webots has been started successfully.
 
-%figure "Security and Privacy settings"
+### Mac OS X security
 
-![mac-unidentified-developper-dialog.png](images/mac-unidentified-developper-dialog.png)
-![mac-security-and-privacy-settings.png](images/mac-security-and-privacy-settings.png)
+During the first launch of Webots, Mac OS X may complain about the opening of Webots because it is from an unidentified developer
```
vincentdm05: "During the first Webots launch" or "During the first launch" is better.
Also, it is not Mac OS X anymore I believe, but rather macOS, so it might be best to call it computer.
"... may complain about opening Webots because ..." is better.

guide/installation-procedure.md
vincentdm05: "this last operation may not have worked." is better.

guide/installation-procedure.md
fabienrohrer: We are using "Mac OS X" everywhere else, so for homogeneity, I would prefer to keep it as-is for now, and come back on this subject in another PR.

Ok for the other modifications.

guide/installation-procedure.md
vincentdm05: Or maybe not, sorry. 👍

vincentdm05: Earlier means older in this case, it's correct ;)

guide/installation-procedure.md
fabienrohrer: Ok, it would be even better to do the contrary. right click is not enabled by default on mac.

guide/installation-procedure.md
vincentdm05: It did work for me on both 10.7 and Mavericks.

omichel: It works at least on Yosemite (I just tested) and I seem to remember it used to work also with earlier versions. So, it is likely work also with El Capitan.

guide/installation-procedure.md
vincentdm05: It is on the pad by double-finger tap.

fabienrohrer:
👍

omichel:
The `Open` menu item should start with a capital letter.

fabienrohrer:
👍

PR #234, opened 2016-11-10
Fix "notes" case

### LINKED ISSUE #189: Consistency of note format
In the User Guide we use the **Note:** format but in the other books (Reference Manual, automobile and darwin-op) we use **note:**.

Personally, I don't have any preference on the format but it would be good to use the same format in all the books.

fabienrohrer:
Fix #189

The "notes" case is incoherent between the `guide | robotis-op2` and the `reference | autombile` books.


fabienrohrer:
👍

PR #239, opened 2016-11-24
Fix cross reference in motor.md

stefaniapedrazzi:
Result: https://www.cyberbotics.com/doc/reference/motor?version=fix-motor-cross-reference


stefaniapedrazzi:
:+1:

PR #240, opened 2016-11-24
Improve wb_display_draw_text() and wb_display_set_font() documentation

fabienrohrer:
Related with https://github.com/omichel/webots/pull/4616

### DIFF
--- a/reference/display.md
+++ b/reference/display.md
@@ -152,7 +152,7 @@ formula.
 
 %end
 
-`wb_display_set_font()` defines the font and the size (in pixel) used for the characters drawn with the `wb_display_draw_text` function, the `anti_aliasing` argument defines whether anti-aliasing filtering should be used to render the characters. The following standard fonts are available:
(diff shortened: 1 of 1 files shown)


fabienrohrer:
👍

reference/display.md
```diff
@@ -247,9 +248,8 @@ closed. Here is an example :
   wb_display_draw_polygon(display,px,py,4); // draw a diamond
 ```
 
-`wb_display_draw_text()` draws an ASCII text from the (`x`,`y`) coordinate. The
-font used to display the characters has a size of 8x8 pixels. There is no extra
-space between characters.
+`wb_display_draw_text()` draws an ASCII text from the (`x`,`y`) coordinate.
+The used font is defined by the [`wb_display_set_font()`](#wb_display_set_color) function.
```
omichel: Shouldn't we say "the font used for drawing the text" rather than "the used font"?

fabienrohrer:
👍

PR #241, opened 2016-11-25
Explain where local licenses are stored

omichel:
I also moved the local license explanation in the license system section which seems more appropriate.


omichel:
👍

PR #245, opened 2016-11-30
Fix wrong italic font in motor feddback section

stefaniapedrazzi:
Fix multiplication sign `*` interpreted as italic section start.


stefaniapedrazzi:
:+1:

PR #246, opened 2016-11-30
Fix emply table header cells in Reference Manual

stefaniapedrazzi:
Empty cells are not correctly displayed in the Webots documentation (but works in the GitHub preview)


stefaniapedrazzi:
:+1:

PR #251, opened 2016-12-12
Update doc to match new default font

tn12787:
This updates the documentation to match the changes in fonts used for rendering text onto Display nodes.


tn12787:
👍

PR #252, opened 2016-12-12
Update default font, remove unused font

tn12787:
This update reverts the default font update in the documentation concerning rendering text onto Display nodes.


tn12787:
👍

PR #259, opened 2016-12-22
Fix glsl variable table

vincentdm05:
Fixes some inaccuracies in the comparison between VRML fields and GLSL variables.


vincentdm05:
👍

PR #263, opened 2016-12-29
Use fish eyes instead of biological eyes

omichel:
This terminology is more common in the literature to designate a spherical projection.


omichel:
👍

tn12787:
👀

PR #264, opened 2017-01-05
Fix OSM importer example

fabienrohrer:
Reference: https://github.com/omichel/webots/blob/master/projects/automobile/resources/OSM_importer/utils.py#L14


fabienrohrer:
👍

tn12787:
👀

fabienrohrer:
> It seems that a reference to the wrong argument remains in the webots repository:
https://github.com/omichel/webots/blob/master/projects/automobile/resources/OSM_importer/importer.py#L52

Well spotted, I'm fixing it in a PR on the webots repo.

PR #266, opened 2017-01-06
Fix spotlight falloff function documentation

vincentdm05:
The documentation now follows the spotlight falloff behaviour.


vincentdm05:
👍

PR #278, opened 2017-02-01
Update Road.proto documentation

fabienrohrer:
- [x] Add missing fields in the Road.proto documentation
- [x] Add the new `turnLane` field: https://github.com/omichel/webots/pull/4830


fabienrohrer:
👍

fabienrohrer:
> Just for curiosity: in which case a road should cast shadows?

I was wondering the same :-)
=> A road could be elevated (for example to mimics a bridge). In this case casting the road shadows may be interesting.

PR #281, opened 2017-02-01
Add instruction to install on linux server edition

stefaniapedrazzi:
Related with https://github.com/omichel/webots/pull/4765.

### DIFF
--- a/guide/installation-procedure.md
+++ b/guide/installation-procedure.md
@@ -139,9 +139,11 @@ gdebi webots_{{ webots.version.major }}.{{ webots.version.minor }}.{{ webots.ver
 
 #### Server edition
 
-Webots requires some graphical features that are usually not available by default on a linux server edition and additional packages needs to be available to make it work.
+Webots requires some graphical features that are usually not available by default on a linux server edition and additional packages needs to be available to make it work:
+* `xserver-xorg-core`
+* `libpulse0`
 
-Webots needs the X Window system and thus the `xserver-xorg-core` and `libxslt1.1` packages have to be installed. These packages are automatically installed when using the DEB package, but in case of the tarball package the user has to manually install them.
+These packages are automatically installed when using the DEB package, but in case of the tarball package the user has to manually install them.
 
 Additionally, it is also necessary to install an OS GUI, for example the Unity desktop `ubuntu-desktop` package.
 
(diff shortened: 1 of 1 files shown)


stefaniapedrazzi:
:+1:

guide/installation-procedure.md
```diff
@@ -137,6 +137,14 @@ or
 gdebi webots_{{ webots.version.major }}.{{ webots.version.minor }}.{{ webots.version.bugfix }}_amd64.deb
 ```
 
+#### Server edition
+
+Webots requires some graphical features that are usually not available by default on a linux server edition and additional packages need to be available to make it work.
+
+Webots needs the X Window system and thus the `xserver-xorg-core` and `libxslt1.1` packages have to be installed. These packages are automatically installed when using the DEB package, but in case of the tarball package the user has to manually install them.
```
omichel: Note that libxslt is not related with the X Window system, it's an XML-related library. Therefore, we shouldn't mention it here.

stefaniapedrazzi:
:+1:

guide/installation-procedure.md
```diff
@@ -137,6 +137,14 @@ or
 gdebi webots_{{ webots.version.major }}.{{ webots.version.minor }}.{{ webots.version.bugfix }}_amd64.deb
 ```
 
+#### Server edition
+
+Webots requires some graphical features that are usually not available by default on a linux server edition and additional packages need to be available to make it work.
+
+Webots depends on `xserver-xorg-core` and `libxslt1.1`. These packages are automatically installed when using the DEB package, but in case of the tarball package the user has to manually install them.
```
omichel: I don't understand why you mention `libxslt1.1` and not all the other dependencies... Can't we include libxslt1.1 also in the tarball, like other dependencies? Why shall we make a special case for this one and not the others?

stefaniapedrazzi: Because this is a special case and only very few user will user a server edition where these two packages are not available.
In desktop edition this packages are installed by default.
It seems a pity to make the tarball package bigger just for very few users.
But I can include them in the tarball package.

omichel: How big is it?

stefaniapedrazzi: I will check in the server edition if copying libxslt is enough or other libraries are required too.
Then I can evaluate the size.

stefaniapedrazzi:
:+1:

omichel:
:eyes:

PR #285, opened 2017-02-08
Remove 'field' word from nodes declaration

stefaniapedrazzi:
Currently the nodes declaration is not consistent: some nodes use the 'field' world but most of them don't use it.

I removed it to make the declaration consistent.
It remains to decide what to do with 'hiddenField' words.

### DIFF
--- a/reference/inertialunit.md
+++ b/reference/inertialunit.md
@@ -4,7 +4,7 @@ Derived from [Device](device.md).
 
 ```
 InertialUnit {
-  MFVec3f    lookupTable    []    # interpolation
+  MFVec3f    lookupTable    [ ]   # interpolation
   SFBool     xAxis          TRUE  # compute roll
   SFBool     zAxis          TRUE  # compute pitch
   SFBool     yAxis          TRUE  # compute yaw
--- a/automobile/ackermannvehicle.md
+++ b/automobile/ackermannvehicle.md
@@ -30,7 +30,7 @@ AckermannVehicle {
   SFFloat    suspensionRearSpringConstant   100000
   SFFloat    suspensionRearDampingConstant  4000
   SFFloat    wheelsDampingConstant          5
-  MFNode     extensionSlot                  NULL
+  MFNode     extensionSlot                  [ ]
   SFNode     boundingObject                 NULL
   SFNode     physics                        NULL
   SFNode     wheelFrontRight                AutomobileWheel { }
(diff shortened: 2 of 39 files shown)


omichel:
That was inspired from VRML97: https://www.evl.uic.edu/swami/VRML97/spec.DIS/part1/nodesRef.html
The idea was to have the `field` or `hiddenField` keyword in front of field names.

I am now wondering whether we could have a more compact representation: removing the `field` keyword and adding a dot prefix to hidden fields (like hidden files on Linux).

stefaniapedrazzi:
Since the documentation is not automatically built from the WRL files, we can temporary keep the same format in WRL as now given that it works properly (low priority).

But in the documentation we can use a more compact representation and the word 'field' is not needed for the users.
Using a dot instead of 'hiddenField' in the documentation is fine for me.

stefaniapedrazzi:
:+1:

reference/inertialunit.md
```diff
@@ -4,7 +4,7 @@ Derived from [Device](device.md).
 
 ```
 InertialUnit {
-  MFVec3f    lookupTable    []    # interpolation
+  MFVec3f    lookupTable    [ ]   # interpolation
   SFBool     xAxis          TRUE  # compute roll
   SFBool     zAxis          TRUE  # compute pitch
   SFBool     yAxis          TRUE  # compute yaw
```
omichel: Note: the spacing is not consistent between the data type and the field name and between the field name and the default value and between the default value and the comment. Sometimes, we have a minimum amount of spaces (up to 1) and sometimes, we have an arbitrary number of spaces, like here 4 or 5.

stefaniapedrazzi: Yes, I notice it too and there is no rule used to decide the spaces that is why I kept them as they are.

omichel:
Also, I have some doubts about the dot notation for the hidden field. I am afraid that users may try to enter the dot in hidden fields in .wbt files and that won't work (unless we support it we could be a good idea...). At least it should be explained somewhere...

reference/inertialunit.md
omichel: We should fix that. Maybe in a different PR.

stefaniapedrazzi: First we should define how to fix it. I will open a new issue.

stefaniapedrazzi:
I remove the dot and instead wrote in the description that the field is hidden.

stefaniapedrazzi:
:+1:

omichel:
:eyes:

PR #291, opened 2017-02-13
Enhancement spaces in nodes declaration (develop)

### LINKED ISSUE #286: Fix spaces in nodes declaration
The spacing is not consistent between the data type and the field name and between the field name and the default value and between the default value and the comment. Sometimes, we have a minimum amount of spaces (up to 1) and sometimes, we have an arbitrary number of spaces, like here 4 or 5.

stefaniapedrazzi:
Fix #286.


stefaniapedrazzi:
:+1:

PR #303, opened 2017-03-08
Fix remove filtering warning doc update

vincentdm05:
Update documentation to mention silent upper bound and max supported af level.

### DIFF
--- a/reference/imagetexture.md
+++ b/reference/imagetexture.md
@@ -57,7 +57,8 @@ clamped in the *s* direction to lie within the [0.0,1.0] range. The `repeatT`
 field is analogous to the `repeatS` field.
 
 The `filtering` field defines the level of quality obtained from filtering
-techniques applied to the texture. This parameter ranges from 0 to 5, 0
+techniques applied to the texture. This parameter ranges from 0 to an integer
+increment relative to the maximum supported anisotropy level (usually 5), 0
 corresponding to no filtering, only performing simple nearest-neighbor pixel
 interpolation. At 1, simple mipmapping is applied. From 2 onwards additional 
 anisotropic filtering is applied with an increasing sampling factor. Using filtering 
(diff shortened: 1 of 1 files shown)


vincentdm05:
👍

reference/imagetexture.md
```diff
@@ -57,8 +57,11 @@ clamped in the *s* direction to lie within the [0.0,1.0] range. The `repeatT`
 field is analogous to the `repeatS` field.
 
 The `filtering` field defines the level of quality obtained from filtering
-techniques applied to the texture. This parameter ranges from 0 to 5, 0
-corresponding to no filtering, only performing simple nearest-neighbor pixel
+techniques applied to the texture. This parameter ranges from 0 to an integer
+increment relative to the maximum supported anisotropy level, which is usually 5.
```
omichel: We should make it clear that 5 is the maximum value.

vincentdm05: It's not always the maximum value, only in the case where it is related to the max supported level.

vincentdm05: In general the max value is defined by the max supported level of the hardware.

vincentdm05:
👍

reference/imagetexture.md
```diff
@@ -57,8 +57,11 @@ clamped in the *s* direction to lie within the [0.0,1.0] range. The `repeatT`
 field is analogous to the `repeatS` field.
 
 The `filtering` field defines the level of quality obtained from filtering
-techniques applied to the texture. This parameter ranges from 0 to 5, 0
-corresponding to no filtering, only performing simple nearest-neighbor pixel
+techniques applied to the texture. This parameter ranges from 0 to an integer
+increment related to the maximum supported anisotropy level. This value usually is 5.
```
omichel: For the sake of clarity I would use "value" instead of "increment" here.

PR #308, opened 2017-03-14
Extend supervisor api

tn12787:
update supervisor doc to show new function,


tn12787:
👍

stefaniapedrazzi:
:eyes:

reference/supervisor.md
```diff
@@ -81,6 +81,7 @@ scored.
 
 WbNodeRef wb_supervisor_node_get_from_def(const char *def);
 WbNodeRef wb_supervisor_node_get_from_id(int id);
+const char *wb_supervisor_node_get_def(WbNodeRef node);
```
stefaniapedrazzi: I will move it one line up (just after `wb_supervisor_node_get_from_def`) instead of splitting `wb_supervisor_node_get_from_id` from `wb_supervisor_node_get_id`.

Same commet for line 75 and 118.

reference/ros-api.md
```diff
@@ -347,6 +347,7 @@ for more information about the value of `robot_unique_name`).
...
 | [/supervisor/node/get\_type\_name](supervisor.md#wb_supervisor_node_get_type)                                      | service       | webots\_ros::node\_get\_type\_name                   | uint64 node<br/>---<br/>string name                                                                                                                                                                                                                               |
+| [/supervisor/node/get\_def](supervisor.md#wb_supervisor_node_get_def)                                              | service       | webots\_ros::node\_get\_def                          | uint64 node<br/>---<br/>string name                                                                                                                                                                                                                               |
```
stefaniapedrazzi: Here now you should use `webots_ros::node_get_name`, the other two `services get_type_name` and `get_base_type_name` have to be updated too.

reference/java-api.md
```diff
@@ -489,7 +489,8 @@ The following tables describe the Java classes and their methods.
...
+| &nbsp;&nbsp; public String [getId](supervisor.md#wb_supervisor_node_get_def)();                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               |
```
stefaniapedrazzi: `[getId]` -> `[getDef]`

reference/cpp-api.md
```diff
@@ -472,7 +472,8 @@ The following tables describe the C++ classes and their methods.
...
+| &nbsp;&nbsp; std::string [getId](supervisor.md#wb_supervisor_node_get_def)() const;                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                            |
```
stefaniapedrazzi: `[getId]` -> `[getDef]`

stefaniapedrazzi: Personally I won't add the new `getDef` function in between `getTypeName()` and `getBaseTypeName` but before `getType` or after `getBaseTypeName` in order to keep semantical groups of functions.

The same comment applies to all the APIs.

reference/supervisor.md
tn12787: I feel it's better here, the first two are about getting nodes based on def/id, the following two (get_def and get_id) are about getting the def/ids themselves. I feel like the current layout makes more sense.

tn12787: Especially considering the way the doc paragraphs are arranged. (line 118)

stefaniapedrazzi: I don't agree.

It is important to keep semantic groups.
So if you say that you want to group all the functions that returns a node, then getId and getDef have to be moved at the end after the other function like getParentNode, getRoot that are returning nodes as well.

Otherwise you have to group the functions related to def name, to the id, etc. mixing sematic groups is bad and makes it difficult to read.

stefaniapedrazzi: Another option is to move getId and getDef functions in the same section than getTypeName.

reference/matlab-api.md
```diff
@@ -513,6 +513,7 @@ The following tables describe the Matlab functions.
 | node = [wb\_supervisor\_node\_get\_self](supervisor.md#wb_supervisor_node_get_from_def)()                                                                            |
 | node = [wb\_supervisor\_node\_get\_from\_def](supervisor.md#wb_supervisor_node_get_from_def)('def')                                                                  |
 | node = [wb\_supervisor\_node\_get\_from\_id](supervisor.md#wb_supervisor_node_get_from_def)('id')                                                                    |
+| s  = [wb\_supervisor\_node\_get\_def](supervisor.md#wb_supervisor_node_get_from_def)(node)                                                                           |
```
omichel: Wrong tag: `#wb_supervisor_node_get_from_def` should be `#wb_supervisor_node_get_def`.

stefaniapedrazzi: Note that it is fine to use the `#wb_supervisor_node_get_from_def` (as it is done for all the other functions in the section), as it links to the correct section.

tn12787:
The ordering in each API's doc is much more consistent now.

👍

PR #311, opened 2017-03-28
Add Supervisor function to restart a Robot's controller

tn12787:
ditto


tn12787:
👍

reference/supervisor.md
```diff
@@ -466,6 +466,25 @@ refer to [this section](#wb_supervisor_simulation_reset_physics).
...
+```c
+#include <webots/supervisor.h>
+
+void wb_supervisor_node_reset_controller(WbNodeRef node);
+```
+
+**Description**
+
+The `wb_supervisor_node_reset_controller()` function restarts the controller of the Robot
+passed to it. If a node other than a Robot is passed to this function, it returns an error.
```
fabienrohrer: > "it returns an error."

This function returns void, so it cannot return an error. However a warning might be displayed in the Webots console.

tn12787:
👍

omichel:
I would call this function `wb_supervisor_node_restart_controller` instead of `wb_supervisor_node_reset_controller`. That seems to be a more precise terminology.

tn12787:
For me restart and reset are equally meaningful and precise here, we are starting the same controller again, this can be viewed as being a reset or a restart, depending on which way you look at it.

omichel:
I believe restart makes a stronger reference to re-starting a process, e.g., the Linux command "service apache2 restart" makes it clear that the `apache2` service will be shutdown and started again. Something like "service apache2 reset" would be less clear to me. In my view, the very same applies to a Webots controller: we are actually restarting its process.

tn12787:
👍

tn12787:
I will merge this PR on @omichel 's approval of the Webots PR.

PR #313, opened 2017-04-03
Add instructions to install Python OpenCV

stefaniapedrazzi:
Related with https://github.com/omichel/webots/pull/4960.

TODO:
* [x] linux instructions
* [x] Windows instructions
* [x] macOS X instructions

### DIFF
--- a/guide/using-python.md
+++ b/guide/using-python.md
@@ -84,13 +84,13 @@ sudo apt-get install python-opencv
 Since macOS El Capitan, open a Terminal and type:
 
 ```
-sudo pip install opencv-python --user
+pip install opencv-python --user
 ```
 
 On older macOS versions, type:
 
 ```
-sudo pip install opencv-python
+pip install opencv-python
 ```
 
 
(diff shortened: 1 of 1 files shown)


stefaniapedrazzi:
@fabienrohrer could you please add the instructions to install python OpenCV on macOS?

fabienrohrer:
`python-opencv` cannot be installed easily on macOS because it seems not to be a member of pip. Using `homebrew` seems to be the official way to do this, but I'm hesitating because of the high risk to override the Python system.

stefaniapedrazzi:
I saw in the text you wrote that is was correct, but not in the comment:
the pip package is called `opencv-python` (Ubuntu 16.04 and Windows), and `python-opencv` is the APT package to be installed on Ubuntu 14.04.

Note that with pip it seems that OpenCV 3 is installed, but this should not be a problem.

fabienrohrer:
Finally, this was an issue about sudo/user in my system. This is now working with these instructions.

fabienrohrer:
👍

stefaniapedrazzi:
:+1:

PR #330, opened 2017-05-05
Sumo configuration file namming convention

DavidMansolino:
Added information about SUMO and NETCONVERT configuration files naming convention (David).

Linked with omichel/webots#5189


DavidMansolino:
👍

PR #332, opened 2017-05-09
Document new FastTree PROTO

DavidMansolino:
Linked with omichel/webots#5168

### DIFF
--- a/automobile/nature.md
+++ b/automobile/nature.md
@@ -50,6 +50,16 @@ randomly selected in the list.
 - `enableBoundingObject`: Defines if the tree should have a bounding object (the
 bounding object is made of a simple cylinder).
 
+### FastTree
+
+The `FastTree` PROTO is similar to the `SimpleTree` PROTO. Iit is less configurable but way faster to load.
+
+#### FastTree Field Summary
+
+- `scale`: The first and last components of the scale define the radius of the tree and the middle one defines it's height.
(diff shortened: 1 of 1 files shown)


DavidMansolino:
👍

automobile/nature.md
```diff
@@ -50,6 +50,16 @@ randomly selected in the list.
 - `enableBoundingObject`: Defines if the tree should have a bounding object (the
 bounding object is made of a simple cylinder).
 
+### FastTree
+
+The `FastTree` PROTO is similar to the `SimpleTree` PROTO. Iit is less configurable but way faster to load.
```
omichel: Typo "Iit".

automobile/nature.md
```diff
@@ -50,6 +50,16 @@ randomly selected in the list.
 - `enableBoundingObject`: Defines if the tree should have a bounding object (the
 bounding object is made of a simple cylinder).
 
+### FastTree
```
omichel: `FastTree` is a strange PROTO name, unlike any other PROTO we have, as it refers to the speed of loading the PROTO and not the resulting object it actually is... I'd rather choose a more neutral name and explain in the documentation that it loads faster than the `SimpleTree` proto. Something like `CustomTree` or maybe, even better, simply `Tree` look better names to me.

DavidMansolino:
👍

PR #343, opened 2017-06-01
Feature supervisor virtual reality headset functions

DavidMansolino:
Linked with omichel/webots#5312


DavidMansolino:
👍

reference/supervisor.md
```diff
@@ -1094,3 +1094,33 @@ an MF\_NODE (like if the node was manually removed from the scene tree).
...
+
+bool          wb_supervisor_virtual_reality_headset_is_used();
+const double *wb_supervisor_virtual_reality_headset_get_position();
+const double *wb_supervisor_virtual_reality_headset_get_orientation();
+```
+
+**Description**
+
+The `wb_supervisor_virtual_reality_headset_is_used()` function returns true if a virtual reality headset is currently used to view the simulation.
+for more information about how to use a virtual reality headset reefer too the [User Guide](https://www.cyberbotics.com/doc/guide/the-user-interface#view-menu).
```
fabienrohrer: Typos here: `For`, `refer`, `to`

DavidMansolino:
👍

reference/supervisor.md
```diff
@@ -1094,3 +1094,33 @@ an MF\_NODE (like if the node was manually removed from the scene tree).
 Note that these functions are still limited in the actual Webots version. For
 example, a device imported into a Robot node doesn't reset the Robot, so the
 device cannot be get by using the `wb_robot_get_device()` function.
+
+---
+
+**Name**
+
+**wb\_supervisor\_virtual\_reality\_headset\_is\_used**, **wb\_supervisor\_virtual\_reality\_headset\_get\_position**, **wb\_supervisor\_virtual\_reality\_headset\_get\_orientation** - *get a handler and more information on a field in a node*
```
omichel: This description seems wrong...

reference/supervisor.md
```diff
@@ -1094,3 +1094,33 @@ an MF\_NODE (like if the node was manually removed from the scene tree).
...
+The `wb_supervisor_virtual_reality_headset_is_used()` function returns true if a virtual reality headset is currently used to view the simulation.
+For more information about how to use a virtual reality headset refer to the [User Guide](https://www.cyberbotics.com/doc/guide/the-user-interface#view-menu).
+
+The `wb_supervisor_virtual_reality_headset_get_position()` and `wb_supervisor_virtual_reality_headset_get_orientation` functions return respectively the current position and orientation of the virtual reality headset as a vector of 3 doubles and a matrix containing 9 doubles that should be interpreted as a 3 x 3 orthogonal rotation matrix:
+```
+[ R[0] R[1] R[2] ]
+[ R[3] R[4] R[5] ]
+[ R[6] R[7] R[8] ]
+```
+If the position or the orientation of the virtual reality headset is not tracked, these functions will return respectively a null vector and an identity matrix.
```
omichel: I would rather return NaN values in that case, because a null vector could be perfecty a legal position.

DavidMansolino: I understand your point, null vector is indeed a perfectly legal position. But it corresponds to the origin, which is equivalent to not tracking the position/orientation. The idea was to avoid that the user has to test for NaN values.

omichel: Yes, but what if I get a (0, 0, 0) vector. Does it mean that tracking is enabled and I am at position (0, 0, 0) or tracking is disabled? How can I distinguish these two cases?

DavidMansolino: My idea was that you don't need to distinguish since applying a null transformation is equivalent to non-tracking the position/orientation. But I agree their might be some cases where you need to know I will change this.

DavidMansolino:
👍

PR #351, opened 2017-06-16
Enhancement to show documentation on robotbenchmark.net

stefaniapedrazzi:
Related with https://github.com/omichel/webots/pull/5394:

Issues:
* [x] specify desired initial menu width
* [x] let the documentation from URL robotbenchmark.net

### DIFF
--- a/js/viewer.js
+++ b/js/viewer.js
@@ -689,20 +689,20 @@ window.onscroll=function(){
 
 
 document.addEventListener("DOMContentLoaded", function() {
-    initializeHandle();
+  initializeHandle();
 
-    if (!setup.url && location.href.indexOf("url=") > -1)
-      setup.url = getGETQueryValue("url", "https://raw.githubusercontent.com/omichel/webots-doc/master/");
-    if (!setup.book)
-      setup.book = getGETQueryValue("book", "guide");
-    if (!setup.page)
-      setup.page = getGETQueryValue("page", "index");
-    if (!setup.anchor)
-      setup.anchor = extractAnchor(location.href);
-    if (!setup.branch)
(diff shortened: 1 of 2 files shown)


stefaniapedrazzi:
:+1:

js/viewer.js
```diff
@@ -8,7 +8,8 @@ if (typeof String.prototype.endsWith !== "function")
         return this.indexOf(suffix, this.length - suffix.length) !== -1;
     };
 
-var local = location.href.indexOf('://www.cyberbotics.com/doc') == -1;
+var local = location.href.indexOf('://www.cyberbotics.com/doc') == -1 &&
+            location.href.indexOf('://robotbenchmark.net/webots_documentation') == -1 ;
```
omichel: It is a pity to hard-code `robotbenchmark.net` here, as we will have to maintain this each time we deploy webots.js on another web site. Can't we find a better solution?

stefaniapedrazzi: It is also a pity that cyberbotics.com is hard coded.
It depends on what is the purpose of this file.
To me it seems much better if the local variable is set in the setup variable. But this would require to change cyberbotics.com

stefaniapedrazzi:
Ok, to keep the cyberbotics website hardcoded given that many functions in `viewer.js`are specific to it.
But I renamed some variables and functions so that they are more self-explanatory.
This changes should be tested on the cyberbotics website, but I don't know how to create a test setup without affecting the one already in production.

omichel:
Unfortunately, there is no test setup for that. We usually try to make it right and immediately test on the production site and revert to the previous version if there is a problem...

stefaniapedrazzi:
:+1:

PR #356, opened 2017-06-20
Fix src/math .gitignore file

fabienrohrer:
Before we change the formulas to MathML, typing `make` in `src/math` is generating annoying log files.


fabienrohrer:
👍

PR #365, opened 2017-06-21
Enhancement automobile scenario creation

DavidMansolino:
I simply added a line about spline subdivison (just to avoid user saying 'the road are not nice after importation').


DavidMansolino:
👍

PR #366, opened 2017-06-21
Drop randomTrips arguments

fabienrohrer:
For all the other SUMO commands, we don't detail superfluous options, because they are specific to the target simulation, and redundant with the SUMO doc. The purpose of this page is to explain only how the scripts and files are interacting and in which order.

Therefore it seems wise to simply remove these cryptic arguments from the doc.


fabienrohrer:
👍

PR #382, opened 2017-07-05
Added a 'name' field to the junction PROTO.

DavidMansolino:
Linked with omichel/webots#5541


DavidMansolino:
👍

automobile/road-segments.md
```diff
@@ -284,6 +284,7 @@ The `Crossroad` PROTO represents a crossroad.
 Crossroad {
   SFVec3f    translation      0 0 0
   SFRotation rotation         0 1 0 0
+  SFString   name             ""                      # could contain the crossroad name
```
omichel: I would remove the comment here to be consistent with the other fields and also because it is repeated below.

DavidMansolino:
👍

PR #387, opened 2017-07-10
Document new Mouse API feature.

DavidMansolino:
Linked to omichel/webots#5575

PR #388, opened 2017-07-10
Feature Supervisor get selected API

DavidMansolino:
Replacement of #387

Linked to omichel/webots#5579


DavidMansolino:
👍

reference/ros-api.md
```diff
@@ -359,6 +359,7 @@ for more information about the value of `robot_unique_name`).
...
 | [/supervisor/get\_from\_id](supervisor.md#wb_supervisor_node_get_from_def)                                            | service       | webots\_ros::supervisor\_get\_from\_id                               | int32 id<br/>---<br/>uint64 node                                                                                                                                                                                                                                  |
+| [/supervisor/get\_selected](supervisor.md#wb_supervisor_node_get_from_def)                                            | service       | webots\_ros::supervisor\_get\_selected                               | uint8 ask<br/>---<br/>uint64 node                                                                                                                                                                                                                                 |
```
stefaniapedrazzi: is not possible to reuse an already existing service?
For example it seems that this one is the same as the one for `wb_supervisor_node_get_from_def`.

I'm not sure what is the best practice in ROS, if reducing the amount of services or having a more specific service name.

reference/supervisor.md
```diff
@@ -137,6 +138,9 @@ The `wb_supervisor_node_get_self()` function returns a handle to the
 utility function that simplifies the task of retrieving the base node without
 having to define a DEF name for it.
 
+The `wb_supervisor_node_get_selected()` function retrieves a handle to the currently selected node.
+It no node is currently selected, the function returns NULL, otherwise, it returns a non-NULL handle.
```
stefaniapedrazzi: It -> If

reference/supervisor.md
```diff
@@ -137,6 +138,9 @@ The `wb_supervisor_node_get_self()` function returns a handle to the
 utility function that simplifies the task of retrieving the base node without
 having to define a DEF name for it.
 
+The `wb_supervisor_node_get_selected()` function retrieves a handle to the currently selected node.
```
stefaniapedrazzi: Do you think that it is better to specify that the selected node refers to the selection in the scene tree?

reference/ros-api.md
tn12787: we already did some service minimization [here.](https://github.com/omichel/webots/pull/4950/files/07165f4c9d64358214dca96c24711d732c6b1cec#diff-f17989f4e1d50b96dd0b8e374b6bf211)

DavidMansolino: It is not the same one as `wb_supervisor_node_get_from_def` has a string as input argument.
We may use the same as for `wb_supervisor_node_get_self` 
but it would look strange to call the `/supervisor/get_selected` service with a `webots_ros::supervisor_get_self` service type. Ideally we should rename the service type `webots_ros::supervisor_get_self` into `webots_ros::supervisor_get_node` (which is more generic) and use it for both service, but that would break the compatibility (maybe we can afford, but it doesn't seems good as we often break compatibility of the ros controller).

stefaniapedrazzi: Sorry, yes I copied the wrong function name.
The point will be to rename the service to something more generic.

stefaniapedrazzi: I agree on the compatibility issue.
The problem is that we never create generic service names and thus we always need a large amount of different service.

But maybe we couldat least  use a generic name for this service without changing the others.

DavidMansolino: I agree, I will rename this service type to `webots_ros::supervisor_get_node` so that it is more generic, and we will be able to use it for other services in the future.

DavidMansolino:
👍

reference/supervisor.md
```diff
@@ -137,6 +138,9 @@ The `wb_supervisor_node_get_self()` function returns a handle to the
 utility function that simplifies the task of retrieving the base node without
 having to define a DEF name for it.
 
+The `wb_supervisor_node_get_selected()` function retrieves a handle to the currently selected node in the scene tree.
+If no node is currently selected, the function returns NULL, otherwise, it returns a non-NULL handle.
+
```
fabienrohrer: I would just add that the selected node is not necessarily a top node.

DavidMansolino:
👍

reference/supervisor.md
```diff
@@ -137,6 +138,9 @@ The `wb_supervisor_node_get_self()` function returns a handle to the
 utility function that simplifies the task of retrieving the base node without
 having to define a DEF name for it.
 
+The `wb_supervisor_node_get_selected()` function retrieves a handle to the currently selected node in the scene tree (which might not be a top node).
+If no node is currently selected, the function returns NULL, otherwise, it returns a non-NULL handle.
```
fabienrohrer: Sounds a little verbose, doesn't it? Could the following be simpler?

```
The `wb_supervisor_node_get_selected()` function returns a handle to the currently selected node in the scene tree.
If no node is currently selected, the function returns NULL.
```

tn12787: I feel like this is more succinct and perfectly describes what the function is for.

DavidMansolino: Actually I mainly copy-pasted this sentence from the description of another function, I will fix both.

DavidMansolino:
👍

tn12787:
I still don't know why we need to say the selected node might not be a top node!

saying we return the selected node is perfectly enough to understand that if i select a node, it'll be correctly returned by this function.

fabienrohrer:
Yes, this is indeed not so logical.

DavidMansolino:
👍

omichel:
I believe we should really get rid of these too specific messages in Webots 8.6.0 even if it breaks the backwards compatibility. We should make this change in a different PR and take are of all those too specific messages in a row. In the future will use only generic messages so that the ROS interface will be and remain clean.

DavidMansolino:
I agree, however I see a few drawbacks of doing this:
  - We will once again break the backward compatibility of the ROS controller, I have the feeling that we are breaking the backward compatibility of this controller at each minor release (we are not allowing ourself to do this with the other API), and this look very unprofessional to me (and may be one of the reason why we see so many users using 'old' version of Webots).
  - If we have general service types, then we will break the matching between Webots API and ROS services (at least for the argument name, here again we are never doing this for other API).
> In the future will use only generic messages so that the ROS interface will be and remain clean

  - It is not always easy to know how to name the service type / argument, when you introduced a new 'generic' message without knowing in advance all its future uses (new API functions).

I will open a new issue to list all the services that could be replaced by more generic ones.

stefaniapedrazzi:
Note that there is already an issue about this topic  #3611

PR #391, opened 2017-07-11
Document new Supervisor import/remove functions.

DavidMansolino:
Linked to omichel/webots#5586


reference/supervisor.md
```diff
@@ -1031,6 +1031,48 @@ and setting fields with the above described functions.
...
+void wb_supervisor_field_import_mf_string(WbFieldRef field, int position, const char *value);
+
+void wb_supervisor_field_remove_mf_bool(WbFieldRef field, int position);
+void wb_supervisor_field_remove_mf_int32(WbFieldRef field, int position);
+void wb_supervisor_field_remove_mf_float(WbFieldRef field, int position);
+void wb_supervisor_field_remove_mf_vec2f(WbFieldRef field, int position);
+void wb_supervisor_field_remove_mf_vec3f(WbFieldRef field, int position);
+void wb_supervisor_field_remove_mf_rotation(WbFieldRef field, int position);
+void wb_supervisor_field_remove_mf_color(WbFieldRef field, int position);
+void wb_supervisor_field_remove_mf_string(WbFieldRef field, int position);
```
DavidMansolino: I am still not sure about these functions, maybe a generic function `wb_supervisor_field_remove_mf` instead of these feild specific one would be better (only for remove, this is not possible for import).

fabienrohrer: I have issues to understand the differences between `wb_supervisor_field_import_mf_bool` and `wb_supervisor_field_set_mf_bool` because they share the same arguments. Could you give an example?

DavidMansolino: `wb_supervisor_field_set_mf_bool` sets the value of the element at 'index' (an element at 'index' should already be present).
`wb_supervisor_field_import_mf_bool` creates a new value and insert it at 'index'.

DavidMansolino: A typical use case of the new function is omichel/webots#5500, currently it is impossible to add new connections to `Crossroads` in the field `connectedRoadIDs` as it is impossible to add new value in a MF field from the Supervisor.

fabienrohrer: In this case, what do you think about replacing the `import` keyword by a list insertion keyword such as `insert`? (inspired from Qt: http://doc.qt.io/qt-5/qlist.html#insert )

When I read the `import` keyword it sounds to me more operations will be performed.

fabienrohrer: The generic remove function seems indeed much better.

DavidMansolino: The idea was to stay close to the MFNode version (wb_supervisor_field_import_mf_node), but in the MF node case it can indeed been seen as an import as a new node is really imported.
Good idea, I will rename `import` into `insert`.

stefaniapedrazzi: Personally I think that the API functions name should be the same for all the field types.
So if you want to use `insert`, then also the MFNode function should have the same name.

DavidMansolino: Unfortunately this would break backward compatibility if we rename the MFNode function.

stefaniapedrazzi: But it doesn't make sense to have different names for MFNode and for the other MF fields.
And this will make the API more complicated for the users.

stefaniapedrazzi: If want to continue with `insert`, then I would deprecate the MFNode import function.

DavidMansolino: I agree, we just need to decide what is the best.
  - Keep import for all of the function.
  - Rename to 'insert' and break backward compatibility.

stefaniapedrazzi: I would go for `insert`.

fabienrohrer: I think the `insert` (the type is directly given) and the `import` (interpreted from strings or files) functions have not the same purpose, so it's justified (and better/simpler in my point of view) to have different keywords.

stefaniapedrazzi: But then why we don't provide a function to insert a node?

We should reduce the number of different functions names if we want that users don't have to search each time in the documentation.

fabienrohrer: Moreover, it's more scalable, we could think to introduce an insert function for the nodes later, taking a WbNodeRef in argument:

```
void wb_supervisor_field_insert_mf_node(WbFieldRef field, int position, WbNodeRef value);
```

WbNodeRef could be created with an API.

fabienrohrer: > But then why we don't provide a function to insert a node?

Because we don't have the API to create WbNodeRef from the controller side.

fabienrohrer: For example, we could think to add later:

```
WbNodeRef new_node = wb_supervisor_create_node("Material");
WbFieldRef color_field = wb_supervisor_node_get_field(new_node, "diffuseColor");
wb_supervisor_field_set_sf_color(color_field, [1, 0, 1]);
wb_supervisor_field_insert_mf_node(somewhere, 12, new_node);
```

stefaniapedrazzi: You can choose the solution you prefer.
But I don't like that MFNode is not handled the same as the other nodes and I don't agree that this is simpler for the users.
At the end they don't care about what we do internally they just want to insert a node in their scene.

DavidMansolino: Maybe we can have a short skype the 3 of us to speak about that?

fabienrohrer: Yes, it would be good.

stefaniapedrazzi: Ok.

DavidMansolino: To summarize our skype discussion:
  - We decided to rename the new functions to `wb_supervisor_field_insert_mf_*`
  - Keep the `wb_supervisor_field_import_mf_node*` function as is (and eventually introduce in the future a `wb_supervisor_field_insert_mf_node` function that would take a `WbNodeRef` as argument, @fabienrohrer will open an issue a bout this).
  - Have a unique `wb_supervisor_field_remove_mf` (or `wb_supervisor_field_remove_mf_item`?) function to remove an item from a MF node.
  - Deprecate the `wb_supervisor_field_remove_mf_node` function.

DavidMansolino:
👍

DavidMansolino:
👍

reference/supervisor.md
```diff
@@ -1031,7 +1031,46 @@ and setting fields with the above described functions.
...
+void wb_supervisor_field_insert_mf_string(WbFieldRef field, int index, const char *value);
+
+void wb_supervisor_field_remove_mf(WbFieldRef field, int index);
+```
+
+**Description**
+
+The `wb_supervisor_field_insert_mf_*()` functions insert a value to a specified multiple `field` (MF).
+The type of the field has to match with the name of the function used, otherwise the value of the field remains unchanged (and a warning message is displayed).
+The `index` parameter defines the position in the MF field where the new node will be inserted.
```
stefaniapedrazzi: new node -> new value

reference/supervisor.md
```diff
@@ -1031,7 +1031,46 @@ and setting fields with the above described functions.
...
+It can be positive or negative. Here are a few examples for the `index` parameter:
+
+- 0: insert at the beginning of the field.
+- 1: insert at the second index.
+- 2: insert at the third index.
+- -1: insert at the last index.
+- -2: insert at the second index from the end of the field.
+- -3: insert at the third index from the end.
+
+The `wb_supervisor_field_remove_mf()` function removes a value to a specified multiple `field` (MF).
```
stefaniapedrazzi: removes a value to -> removes a value from

DavidMansolino:
👍

PR #395, opened 2017-07-17
Update documentation of web streaming

stefaniapedrazzi:
Related to https://github.com/omichel/webots/pull/5615


stefaniapedrazzi:
:+1:

guide/web-streaming.md
```diff
@@ -15,6 +15,19 @@ except that the `X3D` file and the animation are sent on the fly to the Web brow
 
 %end
 
+The web interface displays a toolbar with the following actions:
```
omichel: "actions" sounds strange here. Do you mean "items"?

guide/web-streaming.md
```diff
@@ -15,6 +15,19 @@ except that the `X3D` file and the animation are sent on the fly to the Web brow
...
+* Simulation time: this item indicates the current simulation time on the first line of the simulation element. The second line is currently not used and indicates how long the simulation will run.
```
omichel: I would remove "of the simulation element" which I don't understand... About the second line, it sounds strange: it's not used but still indicates something... We should either remove the fact that it is not used or that it indicates something.

stefaniapedrazzi: The timeout is not used for broadcast streaming so we cannot display any useful value.
Then if we want to reduce the differences with the simulation streaming server .
Many other features are disabled, so this is only another disabled thing and for me it is not something the user will care about.

But you could suggest any other solution.

guide/web-streaming.md
stefaniapedrazzi: No, I really mean actions like play, pause, revert the simulation, etc.
But I will change it with items.

guide/web-streaming.md
omichel: I know. I just wanted to point out that this phrase looks strange. When you say "it's not used and indicates..." seems contradictory. Either it's not used, period, or it's used and it indicates something... But this is not very important...

stefaniapedrazzi: Ok, I understand now. I removed the description and just wrote that it is not used.

stefaniapedrazzi:
:+1:

PR #406, opened 2017-07-25
Custom PROTO fields

tn12787:
linked with https://github.com/omichel/webots/pull/5696

### DIFF
--- a/reference/proto-definition.md
+++ b/reference/proto-definition.md
@@ -83,9 +83,10 @@ IS statements shall refer to fields defined in the PROTO declaration. Multiple
 IS statements for the same field in the PROTO interface declaration is valid.
 
 An interface field that is not linked to an internal PROTO field with an IS will
-produce a warning when parsed. This warning can be suppressed when declaring an 
+produce a warning when parsed. This warning can be suppressed when declaring an
 interface field by using the `unconnectedField` keyword instead of `field` in the
-declaration.
+declaration. This is useful when the value of an interface field is used by some
+Lua code, for example to compute the value of an internal PROTO field.
 
 It is an error for an IS statement to refer to a non-existent interface field.
 It is an error if the type of the field being associated does not match the type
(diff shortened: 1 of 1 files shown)


tn12787:
👍

reference/proto-definition.md
```diff
@@ -82,6 +82,11 @@ IS statements may appear inside the PROTO definition wherever fields may appear.
 IS statements shall refer to fields defined in the PROTO declaration. Multiple
 IS statements for the same field in the PROTO interface declaration is valid.
 
+An interface field that is not linked to an internal PROTO field will produce a
```
omichel: I would add "with a `IS`":
> An interface field that is not linked to an internal PROTO field with a `IS` will produce a

tn12787:
👍

reference/proto-definition.md
```diff
@@ -82,6 +82,11 @@ IS statements may appear inside the PROTO definition wherever fields may appear.
 IS statements shall refer to fields defined in the PROTO declaration. Multiple
 IS statements for the same field in the PROTO interface declaration is valid.
 
+An interface field that is not linked to an internal PROTO field with an IS will
+produce a warning when parsed. This warning can be suppressed when declaring an 
+interface field by using the `unconnectedField` keyword instead of `field` in the
+declaration.
```
omichel: I would add something like: "This is useful when the value of an interface field is used by some Lua code, for example to compute the value of an internal PROTO field."

DavidMansolino: This is not completely correct, as if the field is used by Lua, then the `unconnectedField` is useless as there is no warning anyway.

But I agree it would be nice to explain in which case this keyword is useful. To main the main use is to use this field for 'storing' values used by a Supervisor.

tn12787: I'll add a section about storing data for a supervisor then, that seems like a good use case 😄

tn12787:
👍

PR #416, opened 2017-08-15
Modified tutorial parts 1 and 6.

luc4s:
Part 1 has been changed to match the software behaviour.
Part 6 has been modified to be more concise about the task.


luc4s:
:+1:

PR #429, opened 2017-08-31
Windows: improve instructions to install pyproj

fabienrohrer:
Ditto

### DIFF
--- a/automobile/openstreetmap-importer.md
+++ b/automobile/openstreetmap-importer.md
@@ -32,7 +32,7 @@ pip install lxml pyproj shapely --user
 
 ### Windows
 
-As a prerequisite, `Microsoft Visual C++ Compiler for Python 9.0` or higher is required to build the `pyproj` dependency.
+As a prerequisite, `Microsoft Visual C++ Compiler for Python 2.7` (version `9.0` or higher) is required to build the `pyproj` dependency.
 You can download it [from there](https://www.microsoft.com/en-us/download/details.aspx?id=44266)
 
 Download the shapely wheel called "Shapely‑1.5.17‑cp27‑cp27m‑win_amd64.whl"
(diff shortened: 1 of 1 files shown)


automobile/openstreetmap-importer.md
```diff
@@ -33,6 +33,9 @@ pip install lxml pyproj shapely --user
 
 ### Windows
 
+As a prerequisite, `Microsoft Visual C++ Compiler for Python 9.0` or higher is required to build the `pyproj` dependency.
```
stefaniapedrazzi: This is correct, but I'm wondering if it couldn't be confusing to put the version 9.0 in the name given that on the Microsoft page the software name is `Microsoft Visual C++ Compiler for Python 2.7 `.

fabienrohrer:
👍

automobile/openstreetmap-importer.md
```diff
@@ -33,8 +32,10 @@ pip install lxml pyproj shapely --user
 
 ### Windows
 
-Download the shapely wheel called "Shapely‑1.5.17‑cp27‑cp27m‑win_amd64.whl"
-on http://www.lfd.uci.edu/~gohlke/pythonlibs/#shapely and type a cmd terminal:
+As a prerequisite, `Microsoft Visual C++ Compiler for Python 2.7` (version `9.0` or higher) is required to build the `pyproj` dependency.
+You can download it [from there](https://www.microsoft.com/en-us/download/details.aspx?id=44266).
+
+Download the shapely wheel called [`Shapely‑1.5.17‑cp27‑cp27m‑win_amd64.whl`](http://www.lfd.uci.edu/~gohlke/pythonlibs/#shapely) and type a cmd terminal:
```
omichel: Isn't there a missing word here? The last part of the sentence is hardly understandable. Shouldn't it read "... and type **in** a cmd terminal:". Moreover, I am not sure "cmd terminal" is the right name, I would rather use "cmd.exe command prompt", see https://en.wikipedia.org/wiki/Cmd.exe

tn12787: Or you could even say "in the Windows Command Prompt, `cmd`"

fabienrohrer: You're right, thank you. Just created #431 to fix this.

PR #445, opened 2017-09-20
Fix documentation for the Supervisor API

fabienrohrer:
`wb_supervisor_node_get_number_of_contact_points` returns an `int`, not a `const double *`.


fabienrohrer:
Documentation of this function in the oriented-objects APIs are correct.

fabienrohrer:
👍

PR #446, opened 2017-09-21
Change the return unit of the distance sensor min/max range API

fabienrohrer:
Ditto

### DIFF
--- a/reference/distancesensor.md
+++ b/reference/distancesensor.md
@@ -270,10 +270,12 @@ double wb_distance_sensor_get_aperture(WbDeviceTag tag);
 **Description**
 
 `wb_distance_sensor_get_max_range()` returns the maximum range of the distance
-sensor in meter.
+sensor. This value is computed accorfing to the lookup table of the
+[DistanceSensor](#distancesensor) node.
 
 `wb_distance_sensor_get_min_range()` returns the minimum range of the distance
-sensor in meter.
+sensor. This value is computed accorfing to the lookup table of the
+[DistanceSensor](#distancesensor) node.
 
 The `wb_distance_sensor_get_aperture()` returns the aperture of the distance
 sensor in radians.


fabienrohrer:
👍

omichel:
I am wondering if we could take the opportunity of this PR to rename these functions, so that it's more consistent with the rest of the API and thus more understandable:
```C
double wb_distance_sensor_get_value(WbDeviceTag tag);
double wb_distance_sensor_get_max_value(WbDeviceTag tag);
double wb_distance_sensor_get_min_value(WbDeviceTag tag);
```

fabienrohrer:
It sounds to be a good idea, I'm modifying them here.

reference/distancesensor.md
```diff
@@ -255,25 +255,25 @@ of the return value is defined by this lookup table.
...
+double wb_distance_sensor_get_min_value(WbDeviceTag tag);
 double wb_distance_sensor_get_aperture(WbDeviceTag tag);
 ```
 
 **Description**
 
-`wb_distance_sensor_get_max_range()` returns the maximum range of the distance
-sensor in meter.
+`wb_distance_sensor_get_max_value()` returns the maximum value of the distance
+sensor. This value is the `DistanceSensor.lookupTable[lastVec3f].y` value.
```
stefaniapedrazzi: What if the lookup table is something like this:
[
  0 0 0
  1 100 0
  2 0 0
]

If you return the last value then `wb_distance_sensor_get_max_value` will return 0 and not 100, i.e. it won't really return the max value.

fabienrohrer: I have no solution for this. Do we have such kind of lookupTables?

fabienrohrer: We should rather send the minimum and the maximum value of the WbLookupTable.mOutput.

fabienrohrer: I'm implementing this.

fabienrohrer:
👍

stefaniapedrazzi:
:eyes:

fabienrohrer:
👍

PR #449, opened 2017-09-25
Improved verbosity of local_exporter.py

fabienrohrer:
Aim: typing make on Webots root directory is producing interlaced output. Removing the stdout by default is a solution for this.

- print on stdout only if the --verbose argument is given
- print the errors on stderr.

### DIFF
--- a/local_exporter.py
+++ b/local_exporter.py
@@ -9,6 +9,8 @@ import ssl
 import sys
 import urllib2
 
+verbose = len(sys.argv) and (sys.argv[1] == '--verbose')
+
 dependencies = [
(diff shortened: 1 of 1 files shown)


stefaniapedrazzi:
What about making the inverse, i.e. adding a `--silent` option to be used when called from the main Makefile?
This way by default there will be the verbose output when the script is called directly.

fabienrohrer:
Yes, this sounds better. In this case, I have to modify the main Makefile too.

stefaniapedrazzi:
:eyes:

PR #451, opened 2017-09-25
Merge master to webots 8.6.2

fabienrohrer:
Ditto


fabienrohrer:
(it's arguable to have put automobile/images/titleweb.png in master, because the trucks are not released, but it's minor and the screenshot is better, so I think we can go on).

fabienrohrer:
👍

Issue #254, opened 2016-12-13
Alternative Python installation


fabienrohrer:
We should describe how to use an alternative Python (`make -C resource/languages/python`) in order to support other Python environments (such as the homebrew one)

Slightly related with https://github.com/omichel/webots/pull/4665

fabienrohrer:
.. or Python 3.

This has been required twice this week:
- Cumhur Tuncali : https://www.cyberbotics.com/forum?message=7545
- Kirwan, Ian <ikirwan1@jaguarlandrover.com> : Jaguar / Land Rover

=> I will give it a try today.

fabienrohrer:
Fixed in #455

PR #455, opened 2017-09-28
Document how to use an alternative Python

fabienrohrer:
Attempt to fix: https://github.com/omichel/webots-doc/issues/254

### DIFF
--- a/guide/using-python.md
+++ b/guide/using-python.md
@@ -102,9 +102,7 @@ This documentation is generic and may not cover all the cases.
 This task should be reserved for users having strong knowledge in their OS, build systems
 and in terminal management.
 
-#### Windows
-
-The target Python version for this demo: Python 3.6.
+#### Python 3.6 on Windows
 
 1. Uninstall any Python installation.
 2. Download and install [Python 3.6 for x86_64](https://www.python.org/downloads/):
--- a/guide/using-python.md
+++ b/guide/using-python.md
@@ -97,7 +97,23 @@ where `PYTHON_PATH` is the path to the Python installation directory, for exampl
 
 As described above, the Python libraries for Webots are precompiled with Python 2.7.
 It is possible to use another Python version such as Python 3 or macOS brew Python,
-but this requires to recompile these libraries.
+by recompiling these libraries.
 This documentation is generic and may not cover all the cases.
 This task should be reserved for users having strong knowledge in their OS, build systems
 and in terminal management.
+
+#### Windows
+
+The target Python version for this demo: Python 3.6.
+
+1. Download and install [SWIG precompiled for windows (swigwin)](http://swig.org/download.html).
+2. Download, install and run [MSYS2 for x86_64](http://www.msys2.org/).
+3. From the MSYS2 terminal, type:
+
+```shell
+export PYTHON_VERSION=3.6
+export PYTHON_HOME=/C/Users/$USER/AppData/Local/Programs/Python/Python36
+export PATH=/C/Program\ Files/Webots/msys64/mingw64/bin:/C/Program\ Files/Webots/msys64/usr/bin:$PATH
+cd /C/Program\ Files/Webots/resources/languages/python
+make
+```
--- a/guide/using-python.md
+++ b/guide/using-python.md
@@ -106,19 +106,39 @@ and in terminal management.
 
 The target Python version for this demo: Python 3.6.
 
-1. Download and install [SWIG precompiled for windows (swigwin)](http://swig.org/download.html).
-2. Download, install and run [MSYS2 for x86_64](http://www.msys2.org/).
-3. If Webots is installed in `C:\\Program Files` then the write rights are certainly disabled.
+1. Uninstall any Python installation.
+2. Download and install [Python 3.6 for x86_64](https://www.python.org/downloads/):
+    - Extend your system `PATH` environment variable to add the
+    `C:\Users\fabien\AppData\Local\Programs\Python\Python36` directory.
+    - Add `extern double hypot(double, double);` at the 3rd line of
+    `C:\Users\fabien\AppData\Local\Programs\Python\Python36\include\Python.h`
+3. Download and install [SWIG precompiled for windows (swigwin)](http://swig.org/download.html).
+4. Download, install and run [MSYS2 for x86_64](http://www.msys2.org/).
+5. If Webots is installed in `C:\\Program Files` then the write rights are certainly disabled.
 You should enable the user write rights at least for these paths:
     - `/C/Program\ Files/Webots/resources/languages/python`
     - `/C/Program\ Files/Webots/lib/python`
-4. Remove this file: `/C/Program\ Files/Webots/msys64/usr/bin/swig.exe`.
-5. From the MSYS2 terminal, type:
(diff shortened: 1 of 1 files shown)


fabienrohrer:
Once https://github.com/omichel/webots/pull/6097 merged, I'm able to run our controllers in Python 3 on a clean system, using the instructions described here:

capture d ecran 2017-09-28 a 16 39 42

Unfortunately, these instructions are quite complex.

guide/using-python.md
```diff
@@ -92,3 +92,49 @@ Open the DOS console (CMD.EXE) and type:
...
+This documentation is generic and may not cover all the cases.
+This task should be reserved for users having strong knowledge in their OS, build systems
+and in terminal management.
+
+#### Python 3.6 on Windows
+
+1. Uninstall any Python installation.
+2. Download and install [Python 3.6 for x86_64](https://www.python.org/downloads/):
+    - Extend your system `PATH` environment variable to add the
+    `C:\Users\fabien\AppData\Local\Programs\Python\Python36` directory.
```
omichel: We should add something like `assuming your username is "fabien"`.

guide/using-python.md
```diff
@@ -92,3 +92,49 @@ Open the DOS console (CMD.EXE) and type:
...
+    - Add `extern double hypot(double, double);` at the 3rd line of
+    `C:\Users\fabien\AppData\Local\Programs\Python\Python36\include\Python.h`
+3. Download, install and run [MSYS2 for x86_64](http://www.msys2.org/).
+4. If Webots is installed in `C:\\Program Files` then the write rights are certainly disabled.
+You should enable the user write rights at least for these paths:
+    - `/C/Program\ Files/Webots/resources/languages/python`
+    - `/C/Program\ Files/Webots/lib/python`
+5. Replace every occurrence of `PyString_FromStringAndSize` by `PyString_FromStringAndSize` in
```
omichel: `PyString_FromStringAndSize` are identical...

guide/using-python.md
```diff
@@ -92,3 +92,49 @@ Open the DOS console (CMD.EXE) and type:
...
+`/C/Program\ Files/Webots/resources/languages/python/controller.i`.
+6. Remove the line `init_controller` in `/C/Program\ Files/Webots/resources/languages/python/_controller.def`.
```
omichel: Why? Don't we need it?

guide/using-python.md
```diff
@@ -92,3 +92,49 @@ Open the DOS console (CMD.EXE) and type:
...
+export PYTHON_HOME=/C/Users/$USER/AppData/Local/Programs/Python/Python36
+cd /C/Program\ Files/Webots/resources/languages/python
+make
+```
+
+8. Foreach Python controller, create a `runtime.ini` file containing:
+
+```ini
+[python]
+COMMAND = python
```
omichel: It's a pity... That shouldn't be necessary... We should probably patch Webots, so that it is not needed, e.g., improve the compatibility with Python 3.

omichel:
Instead of that complicated instructions, I am now wondering whether we could simply distribute the MSYS2 terminal with Webots and make it accessible from a Webots menu. Then, we could provide a script that recompile the Python library doing all these complicated things automatically.

fabienrohrer:
I am also not very confident with this solution. I only guarantee it's the simplest in the actual state.

guide/using-python.md
fabienrohrer: This is because Webots tries to run `python27`. I think this is not good, Webots should start `python`. The version check could be done with `python --version`.

omichel:
Note that Anki Cozmo is also apparently using Python 3, see http://cozmosdk.anki.com/docs/install-windows.html, it seems very important to provide a good way to run controllers in Python 3.

omichel:
I believe it is now time to switch from Python 2 to Python 3 by default. See this article for example: https://www.activestate.com/blog/2017/01/python-3-vs-python-2-its-different-time

Webots is on the bleeding edge of R&D and most of the newest deep learning, vision, neural networks and robotics libraries were ported to Python 3 or are being developed in Python 3. Anyhow it should also be possible to recompile for Python 2 following these instructions if needed, but I believe that would be a minority of use cases.

fabienrohrer:
Retarget to `webots8.6.2` branch (because it is requiring the Webots 8.6.2 modifications)

fabienrohrer:
As discussed this morning, the aim is to switch to Python 3 for Webots 2018a, accept this PR (when ready) even if the solution may be complex, and create a `Webots for Python 3 precompiled for Windows` package.

guide/using-python.md
fabienrohrer: My last comment was wrong, this is because a blocking version check is done.
We just discussed with Olivier and decided to let it as-is: in the current situation, it's good to have this Python version check. In the future, we will introduce a global Python command.

guide/using-python.md
fabienrohrer: Swig generates the following lines in:

```
#if PY_VERSION_HEX >= 0x03000000
#  define SWIG_init    PyInit__controller
#else
#  define SWIG_init    init_controller
#endif
```

The entree point differ from Python 2 to 3. I can fix this in the Makefile to simplify the instructions.

fabienrohrer:
Once https://github.com/omichel/webots/pull/6115 is merged:

👍

guide/using-python.md
```diff
@@ -92,3 +92,54 @@ Open the DOS console (CMD.EXE) and type:
...
+
+1. Uninstall any Python installation.
+2. Download and install [Python 3.6 for x86_64](https://www.python.org/downloads/):
+    - Extend your system `PATH` environment variable to add the
+    `C:\Users\$USER\AppData\Local\Programs\Python\Python36` directory.
+    - Add `extern double hypot(double, double);` at the 3rd line of
+    `C:\Users\$USER\AppData\Local\Programs\Python\Python36\include\Python.h`
+3. Download, install and run [MSYS2 for x86_64](http://www.msys2.org/).
+4. If Webots is installed in `C:\\Program Files` then the write rights are certainly disabled.
+You should enable the user write rights at least for these paths:
```
omichel: We could also document that the user could start the MSYS2 console with Administrator rights in order to be able to modify the contents of the following folders. That for example:

"You should either start the MSYS2 console in *administrator* mode (right click on the `MSYS2.exe` icon and select *Run as administrator*) or enable the user write rights at least for these folders:"

guide/using-python.md
```diff
@@ -92,3 +92,54 @@ Open the DOS console (CMD.EXE) and type:
...
+
+
+### Use an alternative Python version
+
+As described above, the Python libraries for Webots are precompiled with Python 2.7.
+It is possible to use another Python version such as Python 3 or macOS brew Python,
+by recompiling these libraries.
+This documentation is generic and may not cover all the cases.
+This task should be reserved for users that have strong experience in their OS,
+build systems and terminal management.
```
omichel: That sounds a bit odd... I would reformulate it simply this way:

"This task requires advanced knowledge in operating systems management."

guide/using-python.md
fabienrohrer: Good idea.

fabienrohrer:
👍

PR #456, opened 2017-10-02
Refactored SUMO interface documentation.

DavidMansolino:
Fixed second part of omichel/webots#6095

https://www.cyberbotics.com/doc/automobile/sumo-interface?version=update-sumo-interface-description

### DIFF
--- a/automobile/sumo-interface.md
+++ b/automobile/sumo-interface.md
@@ -4,24 +4,30 @@ An interface with a microscopic traffic simulator called `SUMO` (Simulation of
 Urban MObility) has been developed using a `Supervisor` node. The advantage of
 interfacing SUMO with Webots is that it allows to easily generate traffic using
 a large number of vehicles in real-time. This interface is written in Python in
-a supervisor controller and uses  [TraCI](http://sumo.dlr.de/wiki/TraCI) to
+a supervisor controller and uses [TraCI](http://sumo.dlr.de/wiki/TraCI) to
 communicate with SUMO.
 
 > **Note**:
 Currently version 0.30 of SUMO is distributed with Webots.
 
-In order to use this interface a few rules need to be observed. First, a
-`Supervisor` node has to be added to the simulation and should use the
-`sumo_supervisor` controller. Then a folder called `worldName_net` should be
-present close to the world file. This folder should contain the usual files
-defining a network in SUMO (.edg.xml, .nod.xml, .rou.xml, etc.) and the
-configuration files (.netccfg and .sumocfg). The configuration files called `sumo.netccfg` and `sumo.sumocfg` will be loaded in priority, if those configuration files do not exist, the interface will look for a configuration file with any other name (it is not recommended to have several configuration files for SUMO or NETCONVERT in the same folder as you can't know which one is going to be used).
+In order to use this interface, a `SumoInterface` PROTO node needs to be added to the world.
+And a folder called `worldName_net` should be present close to the world file.
+This folder should contain the usual files defining a network in SUMO (.edg.xml, .nod.xml, .rou.xml, etc.) and the
+configuration files (.netccfg and .sumocfg), for more information please refeer to the [SUMO documentation](http://sumo.dlr.de/wiki/Networks/SUMO_Road_Networks).
+The configuration files called `sumo.netccfg` and `sumo.sumocfg` will be loaded in priority, if those configuration files do not exist, the interface will look for a configuration file with any other name (it is not recommended to have several configuration files for SUMO or NETCONVERT in the same folder as you can't know which one is going to be used).
 
-Then the interface will automatically start SUMO and run it in synchronization
+The interface will automatically start SUMO and run it in synchronization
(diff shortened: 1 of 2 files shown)


DavidMansolino:
👍

DavidMansolino:
👍

automobile/sumo-interface.md
```diff
@@ -4,79 +4,131 @@ An interface with a microscopic traffic simulator called `SUMO` (Simulation of
...
+```
 
+#### SumoInterface fields summary
+
+- `gui`: Defines if the command-line or GUI version of SUMO should be used.
+- `useNetconvert`: Defines if NETCONVERT should be called before launching SUMO, this should be disabled if a `*.net` is already provided.
+- `enableTrafficLights`: Defines if the traffic lights synchronization should be used or not.
+- `enableWheelsRotation`: Defines if the wheels of the vehicles should rotate or not (wheels rotation has an impact on the simulation speed but make the simulation looks more realistic).
+- `maxVehicules`: Defines the maximum number of vehicles added in Webots from SUMO.
+- `radius`: Defines the visibility radius of the vehicles in meters (only vehicles closer to the viewpoint than this radius will be added in Webots). A negative value means that all the vehicles are added.
```
omichel: The sentence in parenthesis sounds odd. I'd say "(if the distance between the viewpoint and a vehicle is greater than this radius, the vehicle will not added in Webots)".

automobile/sumo-interface.md
```diff
@@ -4,79 +4,131 @@ An interface with a microscopic traffic simulator called `SUMO` (Simulation of
...
+- `maximumLateralSpeed`: Defines the maximal lateral speed of any vehicle in meter per second.
+- `maximumAngularSpeed`: Defines the maximal angular speed of any vehicle in radian per second.
+- `laneChangeDelay`: Defines the time spent to change lane (during this period, the position of the vehicle in Webots and SUMO may not be synchronized anymore).
+- `enableHeight`: Defines whether height information should be extracted from the edge name (experimental).
+- `step`: Defines the time step of SUMO in milliseconds (should be bigger or equal to the world time step).
+- `port`: Defines which port SUMO and Webots should use to communicate.
+- `seed`: Defines the seed of the SUMO random number generator. Use `0` for a random seed (`--random` option of SUMO).
```
omichel: I guess 0 corresponds to a time-based random seed, right? If so, that should documented more precisely.

automobile/sumo-interface.md
```diff
@@ -4,79 +4,131 @@ An interface with a microscopic traffic simulator called `SUMO` (Simulation of
...
+- `networkFiles`: Defines the directory where are located the SUMO network files. If this field is empty, the files should be located in the same directory than the world file in a directory called `worldName_net`.
```
omichel: &rarr; Defines the directory where the SUMO network files are stored.
&arra; If this field is empty, the files should be stored in a directory called `worldName_net` located in the same directory as the world file.

automobile/sumo-interface.md
```diff
@@ -4,79 +4,131 @@ An interface with a microscopic traffic simulator called `SUMO` (Simulation of
...
+- `sumoArguments`: Defines additional arguments passed to SUMO when started.
+- `display`: Can optionally contain a `SumoDisplay` node.
+- `verbose`: Defines if SUMO outputs should be printed in the Webots console.
```
omichel: outputs &rarr; output (this word is always singular in English)

automobile/sumo-interface.md
```diff
@@ -4,79 +4,131 @@ An interface with a microscopic traffic simulator called `SUMO` (Simulation of
...
+]
+```
+
+#### SumoDisplay fields summary
+
+- `width`: Defines the width of the `Display`.
+- `height`: Defines the height of the `Display`.
+- `zoom`: Defines the zooming factor that will make SUMO automatically zoom in/out at startup.
+- `refreshRate`: Defines the refresh rate of the display in milliseconds.
+- `fitSize`: If the current size of the SUMO window is bigger than the resolution defined by the `width` and `height` fields, only the center of the view will be visible in the `Display`. On the opposite, if the SUMO window is smaller than the resolution, the image will not entirely fill it. If `fitSize` is set to `TRUE`, the image will be automatically rescaled, in that case the width / height aspect ratio may not be respected depending on the SUMO window size.
```
omichel: On the opposite &rarr; On the contrary

automobile/sumo-interface.md
DavidMansolino: Yes it is, but on our side what we exactly do is only to send the `--random` argument to SUMO at startup, then SUMO internally uses time-based randomness, but I was not sure we want to document this as this should be documented in the SUMO documentation and maybe not in our.

omichel: Yes, but that sentence sounds wrong, because when we use `0`, it is not a random seed, but rather a time-based seed (non reproducible).

DavidMansolino:
:+1:

PR #464, opened 2017-10-11
Fixed display of include file

omichel:
`<math.h>` file was not displayed on the web page because it was considered as an HTML tag.


omichel:
👍

PR #467, opened 2017-10-18
Update code to support new Webots versioning format R2018 (revision X)

stefaniapedrazzi:
Related to https://github.com/omichel/webots/pull/6182.


stefaniapedrazzi:
With this new method it is no longer possible to build the Webots name ("8", "8.6" or "8.6.2") directly in the documentation because the revision is not displayed if value is 0 and then it is abbreviated to "R2018a_rev1" in the package name.
This is why I think it is much simpler to store directly the three different strings that will be automatically generated from the new version script.

So we will have something like this:

* R2018a (revision 0)
https://github.com/omichel/webots-doc/blob/67e4a28d3acae799a6e650b101f0bcf79fd762da/js/showdown-extensions.js#L29-L35

* R2018a revision 1
```
webots : {
  version : {
    major : "R2018a",
    revision : "R2018a revision 1",
    package : "R2018a_rev1"
  }
}
```

stefaniapedrazzi:
:+1:

omichel:
👀

stefaniapedrazzi:
Yes, the names can be easily changed at any time.
But first the important thing to know is if in general the new behavior introduce here and in https://github.com/omichel/webots/pull/6182 is ok.

stefaniapedrazzi:
I replaced `revision` with `full` and updated the documentation of the Lua `webots_version`.

stefaniapedrazzi:
:+1:

PR #472, opened 2017-10-25
Add description of OpenGL line width preferences option

stefaniapedrazzi:
Related to https://github.com/omichel/webots/pull/6213.

### DIFF
--- a/guide/preferences.md
+++ b/guide/preferences.md
@@ -42,9 +42,9 @@ the Camera device images. The methods are sorted from the most efficient one to
 the least efficient one.
 
 - The `Main 3D view anti-aliasing` option allows you to enable Multisample Anti-Aliasing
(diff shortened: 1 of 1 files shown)


stefaniapedrazzi:
:+1:

tn12787:
Does this parameter also affect the lines drawn in the translate/rotate handles?

stefaniapedrazzi:
Yes, they are the coordinate system lines.

tn12787:
OK. I think it would be good to explicitly refer to these (along with the force and torque handles).

stefaniapedrazzi:
Ok. I think that they are already referred as coordinate system because this is what they are representing and having a too long and too specific list is just worst.
But I can add them.

stefaniapedrazzi:
:+1:

PR #478, opened 2017-10-26
Improve charger.png picture

### LINKED ISSUE #477: Fix Charger radius pictures
The pictures describing the Charger radius field are misleading:
https://www.cyberbotics.com/doc/reference/charger#the-sensitive-area-of-a-charger

From this picture it seems that if you add a Transform into the Charger node children field, then you can move the charging area.
But it is wrong: the center of the charging area always corresponds to the center of the Charger node.

Personally I would complete remove the second case picture and improve the text and labels of the first one.

Note that the text is correct.

stefaniapedrazzi:
Fix #477: remove second case and unneeded text.


stefaniapedrazzi:
:+1:

stefaniapedrazzi:
:-1: 

I'm also checking if the description of the other fields is correct.

stefaniapedrazzi:
:+1:

PR #489, opened 2017-11-13
Remove description of unrelated fields in JointParameters

stefaniapedrazzi:
`minPosition`and `maxPosition`are `Motor` fields and not `JointParameters` fields.


stefaniapedrazzi:
:+1:

PR #495, opened 2017-11-15
Fixed shapely installation instruction on windows

DavidMansolino:
ditto


DavidMansolino:
👍

DavidMansolino:
Note sure this is really useful as only one version is available on the page.

PR #499, opened 2017-11-16
Feature animation/streaming update rate

DavidMansolino:
Dual PR of omichel/webots#6347


DavidMansolino:
👍

guide/web-animation.md
```diff
@@ -45,13 +45,32 @@ Please refer to [this section](web-scene.md#how-to-embed-a-web-scene-in-your-web
...
 The other VRML fields are not recorded in the animation file.
 Node insertion or deletion are also not recorded in the animation file.
 
 Additional limitations are described in [this section](web-scene.md#limitations).
 
 
+### Scene refresh rate
+
+The scene refresh rate is defined by the `WorldInfo.FPS` field.
+However, the `WorldInfo.basicTimeStep` field have an impact on the refresh rate too, because the update are only sent during the simulation step. The actual refresh rate can be computed with the following formula:
```
stefaniapedrazzi: update are only sent -> update**s** are only sent

guide/web-animation.md
```diff
@@ -45,13 +45,32 @@ Please refer to [this section](web-scene.md#how-to-embed-a-web-scene-in-your-web
...
+
+%figure "Animation refresh rate."
```
stefaniapedrazzi: to me it looks a little bit strange to have a caption for a formula

DavidMansolino:
> rate = 1000 / min(step * ceil(1000 / (step * FPS)), step)

Shouldn't it be `rate = 1000 / min(step / ceil(1000 / (step * FPS)), step)` instead?

stefaniapedrazzi:
> Shouldn't it be rate = 1000 / min(step / ceil(1000 / (step * FPS)), step) instead?

No, `(1 /a) / b = 1 / (a * b)`.
And the the `ceil(..)` represents the number of steps so it has to be multiplied by the step value.

stefaniapedrazzi:
By the way I think it is better to use the complete name `basicTimeStep`.
I used step in my comment because it was shorter.
I'm sorry that it was not clear.

DavidMansolino:
> No, (1 /a) / b = 1 / (a * b).

You are absolutely right, I did get confuse. Then maybe a simpler version would be:
codecogseqn 1

(Personnally keeping the (1000 / basicTimeStep) together helps because it represents the world refresh rate). Is this version simpler for you or you prefer:
codecogseqn 2

stefaniapedrazzi:
Yes, your proposal is also good.

DavidMansolino:
👍

PR #503, opened 2017-11-20
remove non-selling items

tn12787:
Removes items which are better suited to the changelog and don't really "show off" Webots 2018a.


tn12787:
👍

PR #504, opened 2017-11-21
Documentation mistake (Connector)

luc4s:
The documentation page about the connector component states that the y-axis is drawn in red:

> The z-axis is drawn as a 5 cm blue line, the y-axis (a potential docking rotation) is drawn as a 5 cm red line, (...)

 This is not correct and has been modified appropriately (y-axis is actually drawn in green).


luc4s:
:+1:

PR #506, opened 2017-11-22
Add Solid name to PROTO descriptions

stefaniapedrazzi:
Related to https://github.com/omichel/webots/pull/6380: update description of PROTO nodes to include the Solid.name parameter.

### DIFF
--- a/automobile/automobilewheel.md
+++ b/automobile/automobilewheel.md
@@ -14,6 +14,7 @@ PROTO featuring the corresponding `automobileWheel` slots.
 
 ```
 AutomobileWheel {
+  SFString   name               "automobile wheel"
   SFFloat    thickness          0.3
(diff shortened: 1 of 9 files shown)


stefaniapedrazzi:
:+1:

fabienrohrer:
I'm wondering if we should not also add a note about the uniqueness rule of the Solid.name in the definition of name: https://www.cyberbotics.com/doc/reference/solid

stefaniapedrazzi:
> I'm wondering if we should not also add a note about the uniqueness rule of the Solid.name in the definition of name: https://www.cyberbotics.com/doc/reference/solid

Yes, of course. But this has not been implemented yet.
In this first step I'm only adding a Solid.name parameters to the PROTO nodes.

PR #529, opened 2017-12-13
Update automobile python import

DavidMansolino:
ditto


DavidMansolino:
👍

DavidMansolino:
You are right I missed this one, but I think it is: import com.cyberbotics.webots.**controller**.vehicle.Driver

DavidMansolino:
👍

PR #535, opened 2017-12-18
Fix OSM importer dependencies on Ubuntu 14.04

DavidMansolino:
Fix omichel/webots#6540

### DIFF
--- a/automobile/openstreetmap-importer.md
+++ b/automobile/openstreetmap-importer.md
@@ -15,7 +15,7 @@ the importer script.
 
 Follow [these instructions](https://www.cyberbotics.com/doc/guide/using-python) to install Python.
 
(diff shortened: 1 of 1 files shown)


DavidMansolino:
:+1:

stefaniapedrazzi:
:eyes:

stefaniapedrazzi:
On Ubuntu 14.04 it works fine but there is the same issue also on Ubuntu 16.04.

It would be better to test the OSM importer also on macOS.

DavidMansolino:
Actually `webcolors` is a new dependency introduced in omichel/webots#5542 and it should therefore be installed on all the platforms.

DavidMansolino:
:+1:

PR #536, opened 2017-12-18
Fixed the display of branches and tag based documentation versions

omichel:
Address https://github.com/omichel/webots/issues/6539

The current system seems to be broken as:
1. it expects that tags are set properly in webots-doc, which is not the case.
2. it recognizes the version number from the `major.minor.maintenance` template which doesn't work any more with `R2018a`.
3. it uses both the tag and branch concept which is over-complicated for your purposes (see below).

Nothing has to be changed on the Webots side as the links to the online documentation are properly generated with the correct version number, corresponding to the version of Webots currently in use.

The concept of "tag" in the documentation script seems to be useless. Actually we can access raw content of github branches and tags exactly the same way. So, we can simplify our script and store the branch name only (which may also be a tag name).

Also the `getWebotsVersion()` will return the branch/tag name if specified or the current version if not specified, which is exactly what to want to display on the web page.

Unfortunately, it is possible to test these scripts until they are actually in production. I quickly tested them and they appear to behave nicely between Webots 8.5, 8.6 and Webots R2018a.

I will now add the tags for Webots 8.5.

### DIFF
--- a/doc.php
+++ b/doc.php
@@ -15,22 +15,15 @@
     if ($j === FALSE) {
       $page = substr($uri, $i + 1);
       $branch = '';
-      $tag = '';
     } else {
       $page = substr($uri, $i + 1, $j - $i - 1);
       $version = substr($uri, $j + 9);
-      if (preg_match('/^\d+\.\d*(.)+$/',$version)) {
-        $branch = '';
-        $tag = $version;
-      } else {
-        $n = strpos($version, ':');
-        if ($n === FALSE)
-          $branch = $version;
-        else {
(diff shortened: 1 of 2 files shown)


omichel:
Because the new documentation system was introduced in Webots 8.5.0, there is no need to go further in the tag history of webots-doc. I just added the 8.5 tag and I believe we are fine for now.

omichel:
👍

omichel:
Note: I updated the release procedure on the Wiki to remember to add the version tag on the webots-doc repository.

Issue #402, opened 2017-07-20
TTF fonts on mac


fabienrohrer:
In https://github.com/omichel/webots/pull/5667 I had to workaround a bug in QtWebKit about the support of CSS font-face.

It could be good to check if Qt > 5.9 solves this.

fabienrohrer:
Fixed in https://github.com/omichel/webots-doc/pull/494

omichel:
@fabienrohrer: Can we close this one as well (we don't use Raleway any more)?

fabienrohrer:
I think it would be better to keep it open, it's the only trace we have about the fact that the fonts are buggy on macOS, and that we have a fallback for this.
The long term solution is probably to upgrade to QtWebEngine which hopefully supports better fonts.

fabienrohrer:
Fixed in omichel/webots#7316

PR #537, opened 2017-12-18
Forced an helvetica font on macOS.

fabienrohrer:
https://github.com/omichel/webots-doc/issues/402 is back again. I applied this dirty fix for the release.


fabienrohrer:
👍

PR #541, opened 2017-12-20
Merge develop to master

stefaniapedrazzi:
Ditto.


stefaniapedrazzi:
:+1:

fabienrohrer:
Note: This merge should be applied only when the packages are released on the website.

stefaniapedrazzi:
We should also do it before in order to update the R2018a, otherwise the the online doc will point to an outdated page.

We should find a tradeoff.

PR #542, opened 2017-12-21
"Webots Devices" section refactor

fabienrohrer:
Before: https://cyberbotics.com/doc/guide/webots-devices
After: https://cyberbotics.com/doc/guide/webots-devices?version=update-sensor-samples-section

Aim:

When a user enters on this page, he expects:

- to find easily an example demonstrating the device he would like to work with (keywords, explanations)
- Understand what he can expect from this sample (behavior).

TODO:

- [x] Redo screenshots
- [x] Proofreading existing sub sections.
- [x] Be exhaustive
- [x] Improve keywords


fabienrohrer:
I plan to remove the screenshots of this page (except the generic one on the top, and eventually adding specific or good looking ones, like the pen, the lidar or so), because they are very redundant and do not bring much.

omichel:
Couldn't we replace them with X3DOM animation or scenes instead?

omichel:
I don't agree on removing the screenshot and not replacing them with X3DOM views or animations. I believe an image says 100x more than words... We may reduce their size or improve them in some way, but I believe they are really useful.

fabienrohrer:
I just discussed with Olivier. We will postpone investigations on X3D scenes inclusion later. We will keep images, but smaller, and trying to link them better with the text.

fabienrohrer:
I'm adding a new showdown rule + CSS to deal better with illustrated sections like this:

` In this example, etc.`

It will affect these pages:

```
$ egrep -e '!\[.*\]\(.*\) ' */*.md | sed -e 's/:.*//' | uniq
guide/the-scene-tree.md
guide/the-user-interface.md
guide/web-streaming.md
guide/webots-built-in-editor.md
guide/webots-devices.md
robotis-op2/examples.md
```

fabienrohrer:
I updated the files I touched (JS, CSS, CSS) on the FTP so the online doc is compliant with this modification:

https://cyberbotics.com/doc/guide/webots-devices?version=update-sensor-samples-section

fabienrohrer:
👍

stefaniapedrazzi:
Why did you update the FTP files before the PR is accepted?
Are these files used by the deployed documentation?

fabienrohrer:
Yes.

I have done this with care:
- the CSS modification is touching only the new class.
- the showdown and viewer modification is only about adding a new extension occurring only in this case.

stefaniapedrazzi:
We can discuss this is in the meeting but it is a bad procedure.
Nothing deployed should be changed without at least a second review.

Note that also robotbenchmark depends on the documentation files on the FTP and the released version of Webots and probably you didn't though about testing it.
In this special case we are still using Webots 8.6 in the server machines, but this is just a case.

fabienrohrer:
I fully agree with you, I know I'm breaking our rules there.

You also have to take into consideration that the devices chapter is in a really bad state (it's a really bad advertising for us), and this is the quickest way to solve this. I think rigid rules are necessary, but we should also break them when it is meaningful.

stefaniapedrazzi:
I don't think that waiting for a review would make the difference in terms of time needed to solve the issue...

omichel:
I believe that a solution that relies on https://rawgit.com to refer to our github branch files should be investigated, so that we can review such PRs without any risk of breaking our deployed code.

fabienrohrer:
I'm opening a new issue for this topic.

fabienrohrer:
👍

fabienrohrer:
👍

fabienrohrer:
👍

omichel:
:eyes:

fabienrohrer:
Tom fixed english in 8b8cfae, and I didn't added anything new since.

PR #544, opened 2018-01-03
Improved Car fields description

DavidMansolino:
Added a link to the engine model section which explains in more detail the meaning of these fields.


DavidMansolino:
👍

Issue #327, opened 2017-05-03
Link across books not working


DavidMansolino:
It should be possible to create hyperlinks across books using relative pathes (e.g. from the guide use `[Camera](../reference/camera.md)`). Unfortunately it seems to work on Github but not in our viewer (on Cyberboitics website).

Once fixed, the references in the description of `radarCrossSection` and `recognitionColors` in automobile/ackermannvehicle.md` should be added.

PR #550, opened 2018-01-18
Simulation Reset

DavidMansolino:
Linked to omichel/webots#6640

https://www.cyberbotics.com/doc/guide/the-user-interface?version=feature-fast-revert


omichel:
I would rather call this feature "soft revert" or even better "reset simulation" or simply "reset" instead of "fast revert". What do you think?

DavidMansolino:
👍  'reset' seems indeed a better keyword than 'fast revert'.

tn12787:
I think if we are to call this "reset" then traditional "revert" should be called "reload".

DavidMansolino:
👍

tn12787:
👀

guide/the-user-interface.md
```diff
@@ -46,16 +46,23 @@ filename entered by the user. Note that a ".wbt" file should always be saved in
...
 load a window will pop up and show the progression. Using this window it is
 possible to cancel the loading of the world, in that case, the empty world will
 be opened instead.
 
+- ![](images/reset-simulation-button.png =26x26) The **Reset Simulation** menu item (and button) restore the initial state of the simulation.
+The simulation is not entirely destructed and reconstructed like with a reload, but the initial state of all the nodes is restored, which is way faster.
+
+    > **Note**:
+There are currently some limitations to the reset.
+The node removed from the simulation (either manually or from a Supervisor) are not re-imported during the reset. Furthermore, not all the changes to the node fields are reverted.
```
omichel: &rarr; The node**s**

reference/supervisor.md
```diff
@@ -770,6 +748,35 @@ overwrite silently existing files, so that the corresponding data may be lost.
...
+
+```c
+#include <webots/supervisor.h>
+
+void wb_supervisor_simulation_reset();
+```
+
+**Description**
+
+The `wb_supervisor_simulation_reset()` function sends a request to the simulator process, asking it to reset the simulation immediately.
```
stefaniapedrazzi: This description is exactly the same as the reload world one.
As a user that will read the documentation I won't be able to understand the difference between these two functions.
The only explained difference is that the seed of random number generators is reset.

guide/the-user-interface.md
```diff
@@ -46,16 +46,23 @@ filename entered by the user. Note that a ".wbt" file should always be saved in
...
 from the beginning.
 
     > **Note**:
-When opening (or reverting) a world, if the world takes more than 2 seconds to
+When opening (or reloading) a world, if the world takes more than 2 seconds to
 load a window will pop up and show the progression. Using this window it is
 possible to cancel the loading of the world, in that case, the empty world will
 be opened instead.
 
+- ![](images/reset-simulation-button.png =26x26) The **Reset Simulation** menu item (and button) restore the initial state of the simulation.
```
stefaniapedrazzi: restore -> restore**s** (menu item is singular)

guide/the-user-interface.md
stefaniapedrazzi: Which are the node fields changes that are not reverted?
The user needs to know this information.

DavidMansolino:
👍

reference/supervisor.md
```diff
@@ -772,11 +772,13 @@ void wb_supervisor_simulation_reset();
 **Description**
 
 The `wb_supervisor_simulation_reset()` function sends a request to the simulator process, asking it to reset the simulation immediately.
+The simulation is not entirely destructed and reconstructed like when calling `wb_supervisor_world_reload`, but the initial state of all the nodes is restored, which is way faster.
+This function resets also the seed of the random number generator used in Webots, so that noise-based simulations can be be reproduced identically after calling this function.
```
omichel: be be

stefaniapedrazzi:
Probably it would be good to first synchronize with the master branch, because the way the function names modified in this PR are written has been changed: remove brackets, added function suffix etc.

reference/supervisor.md
```diff
@@ -767,9 +745,40 @@ indicates the success of the save operation. Be aware that this function can
...
+void wb_supervisor_simulation_reset();
+```
+
+**Description**
+
+The `wb_supervisor_simulation_reset()` function sends a request to the simulator process, asking it to reset the simulation immediately.
+The simulation is not entirely destructed and reconstructed like when calling `wb_supervisor_world_reload`, but the initial state of all the nodes is restored, which is way faster.
+This function resets also the seed of the random number generator used in Webots, so that noise-based simulations can be be reproduced identically after calling this function.
+There are currently some limitations to the reset, The nodes removed from the simulation (either manually or from a Supervisor) are not re-imported during the reset.
+Furthermore, changes to node fields done manually or using a Supervisor are not reverted.
```
stefaniapedrazzi: `changes to node fields done manually or using a Supervisor`: so which changes are reverted?

reference/supervisor.md
```diff
@@ -767,9 +745,40 @@ indicates the success of the save operation. Be aware that this function can
...
+
+void wb_supervisor_simulation_reset();
+```
+
+**Description**
+
+The `wb_supervisor_simulation_reset()` function sends a request to the simulator process, asking it to reset the simulation immediately.
+The simulation is not entirely destructed and reconstructed like when calling `wb_supervisor_world_reload`, but the initial state of all the nodes is restored, which is way faster.
+This function resets also the seed of the random number generator used in Webots, so that noise-based simulations can be be reproduced identically after calling this function.
+There are currently some limitations to the reset, The nodes removed from the simulation (either manually or from a Supervisor) are not re-imported during the reset.
```
stefaniapedrazzi: `, The` -> `, the`

reference/supervisor.md
```diff
@@ -767,9 +745,40 @@ indicates the success of the save operation. Be aware that this function can
...
+```c
+#include <webots/supervisor.h>
+
+void wb_supervisor_simulation_reset();
+```
+
+**Description**
+
+The `wb_supervisor_simulation_reset()` function sends a request to the simulator process, asking it to reset the simulation immediately.
+The simulation is not entirely destructed and reconstructed like when calling `wb_supervisor_world_reload`, but the initial state of all the nodes is restored, which is way faster.
```
stefaniapedrazzi: For consistency with the new changes of @fabienrohrer  you should use the function suffix:
" when calling `wb_supervisor_world_reload`" -> " when calling the `wb_supervisor_world_reload` function"

reference/supervisor.md
stefaniapedrazzi: You should better described which changes are reverted and which are not, because it is still not clear.

stefaniapedrazzi: The addition of the `in the scene tree` helps. about the Supervisor part it was already clear before.
But if I would read this sentence for the first time the question I will have is always the same: if all the changes I apply are not reverted which changes are reverted?

I think that there should be a generic rule to define which fields are reverted, otherwise if each field behaves different it doesn't seem user-friendly.

DavidMansolino: Just to summarize and help find a better explanation/generic rule, here is what is happening step by step when doing a simulation reset:
  - Simulation time is set to 0.0
  - All the nodes added during the simulation are removed.
  - All the sound sources are stoped.
  - The random seeds are reset.
  - All the nodes are reset:
    - **Brake**: The brake is released.
    - **Charger**: The battery level is restored (the `battery` **field** is restored), `emissiveColor` *field* of Material is restored.
    - **Connector**: If attached the connector is detached and the value of the `isLocked` **field** is restored.
    - **Emitter-Receiver**: the message queue is cleared.
    - **Joint/Motor**: The position/velocity/torque/force/acceleration is restored.
    - **LED**: The light is switched off. The emissiveColor **field** of the Material is restored.
    - **Lidar**: The position of the rotating head is restored.
    - **Pen**: All the painted texture are cleaned.
    - **Propeller**: The slow helix and it's initial position are restored.
    - **Robot**: The battery level is restored (`battery` **field** is restored). The controller is restarted.
    - **Solid**: The initial translation and rotation are restored (`translation` and `rotation` **fields** are restored) and the physic is reset.
    - **Supervisor**: All the labels are removed.
    - **Track**: the motor position is restored and the translation of the textureTransform is restored (`translation` **field** is restored).
    - **Viewpoint**: the position and orientation **fields** are restored.

The only generic rule I see is: the restored fields are the one changed by the simulation itself and not by an external intervention (manually in the scene tree or by a Supervisor).

stefaniapedrazzi: The generic rule that I see is that the devices and objects internal states modified by the simulation itself and by the controller are reset.

stefaniapedrazzi: But it is very important to write this as clearly as possible in the documentation.

stefaniapedrazzi: Examples are also useful to explain the behavior, you do not have to list everything, but probably mentioning the most important things that are reset will help the user.

DavidMansolino:
👍

guide/the-user-interface.md
```diff
@@ -46,16 +46,41 @@ filename entered by the user. Note that a ".wbt" file should always be saved in
...
+In order to reset the simulation, the following steps are performed:
+  - The simulation time is set to 00:00:00
+  - All nodes added during the simulation are removed.
+  - All sound sources are stopped.
+  - The random seeds used by Webots internally are reset.
+  - All the nodes are reset. This has the following implication depending on the node type:
+    - **Brake**: The brake is released.
+    - **Charger**: The `battery` field and the `emissiveColor` field of the Material node of the first Shape child node are restored.
+    - **Connector**: If attached, the connector is detached and the value of the `isLocked` field is restored.
+    - **Emitter-Receiver**: the message queue is cleared.
```
stefaniapedrazzi: in all the other items you start with a capital letter

reference/supervisor.md
```diff
@@ -767,9 +745,37 @@ indicates the success of the save operation. Be aware that this function can
...
+```c
+#include <webots/supervisor.h>
+
+void wb_supervisor_simulation_reset();
+```
+
+**Description**
+
+The `wb_supervisor_simulation_reset` function sends a request to the simulator process, asking it to reset the simulation immediately.
+The reset process is explained in detail in the [User Guide](https://www.cyberbotics.com/doc/guide/the-user-interface#file-menu).
```
stefaniapedrazzi: Is this link still correct if a specific version is displayed?
For example when opening the online doc for a specific release?

guide/cpp-java-python.md
```diff
@@ -45,7 +45,7 @@ Hence it is possible to call the `Robot`'s methods, such as, e.g., the `step` or
 `getLED` functions, from the `DifferentialWheels` and `Supervisor` controllers. But it
 is not possible to call the `Supervisor` methods from a `DifferentialWheels`
 controller, and vice versa. For example it won't be possible to call
-the `simulationRevert` function from a `DifferentialWheels` controller.
+the `worldReload()` function from a `DifferentialWheels` controller.
```
stefaniapedrazzi: brackets have to be removed

reference/supervisor.md
DavidMansolino: It will work but the `version` will be removed.
As long as #327 is not fixed, I don't see any other way to reference another book.

stefaniapedrazzi: Yes, it will link to the wrong documentation and the possibly wrong explanation of the reset functionality.
This is bad, and we should find a solution.
For me it can be fixed in another issues, but it is important to fix this soon.

DavidMansolino: 👍 
I did set the milestone of the issue to R2018b so that we make sure it is fixed before this become public.

reference/supervisor.md
```diff
@@ -767,9 +745,37 @@ indicates the success of the save operation. Be aware that this function can
...
-In the other APIs, the `Robot.saveWorld` function can be called without
+In the other APIs, the `Robot.worldSave` function can be called without
 argument. In this case, a simple save operation is performed.
 
+The `wb_supervisor_world_reload` function sends a request to the simulator
+process, asking it to reload the current world immediately. As a result of
+reloading the current world, the supervisor process and all the robot processes
+are terminated and restarted. You may wish to save some data in a file from your
+supervisor program in order to reload it when the supervisor controller
+restarts.
```
omichel: This doesn't respect the one sentence per line rule.

robotis-op2/using-remote-control-from-robot-window.md
```diff
@@ -26,9 +26,9 @@ remote control has been sucessfully started.
 You can now easily start and pause
 your controller in remote control mode by using the run and pause buttons of
 Webots simulation (see chapter about [examples](examples.md) for more details).
-Warning: if you revert
-the simulation it will stop the remote control mode. In order to stop the
-remote control (without reverting) simply press the following button:
+Warning: if you reload
+the world it will stop the remote control mode. In order to stop the
+remote control (without reloading) simply press the following button:
```
omichel: One sentence per line.

DavidMansolino: Do we really want to introduced so many changes for non-new sentences, because it will hopefully be done automatically in #587 and introduce a lot of conflicts that we will have to manually manage if we changes all the paragraph that doesn't respect the rule now?

DavidMansolino:
👍

DavidMansolino:
:+1: 

I just fixed conflicts:
  - guide/modeling.md
  - guide/programming.md
  - guide/the-user-interface.md
  - guide/using-the-nao-robot.md
  - guide/using-visual-cpp-with-webots.md
  - reference/camera.md
  - robotis-op2/examples.md

guide/the-user-interface.md
```diff
@@ -28,14 +28,39 @@ On each `Save` the content of the ".wbt" file is overwritten and no backup copie
...
+    - **Charger**: The `battery` field and the `emissiveColor` field of the Material node of the first Shape child node are restored.
+    - **Connector**: If attached, the connector is detached and the value of the `isLocked` field is restored.
+    - **Emitter-Receiver**: The message queue is cleared.
+    - **Joint/Motor**: The position, velocity, acceleration, available torque and available force are restored.
+    - **LED**: If the first child is a Light node, it's `color` field is restored and it is switched off. If the first child is a Shape node, the `emissiveColor` field of its Material node is restored.
+    - **Lidar**: The position of the rotating head is restored.
+    - **Pen**: All the painted textures are cleaned.
+    - **Propeller**: The slow helix and it's initial position are restored.
+    - **Robot**: The `battery` field is restored and the controller is restarted.
+    - **Solid**: The `translation` and `rotation` fields are restored and the physic is reset.
```
omichel: Typo: physic &rarr; physics

PR #566, opened 2018-01-30
Remove unprotected underscores in titles (and test this)

fabienrohrer:
- [x] Remove unprotected underscores in title (and add a test)
- [x] Remove remaining code statements in hyperlinks (and update the test)
- [x] Ignored the pep8 'E501' rule about lines with > 80 character (to match our coding style)

Typical underscore issue:

- https://www.cyberbotics.com/doc/guide/controller-programming#simulation-step-and-wbrobotstep
- **Fixed**: https://www.cyberbotics.com/doc/guide/controller-programming?version=enhancement-underscores-in-title#simulation-step-and-wb_robot_step

Typical link issue:

- https://www.cyberbotics.com/doc/reference/robot#wb_robot_window_custom_function
- **Fixed**: https://www.cyberbotics.com/doc/reference/robot?version=enhancement-underscores-in-title#wb_robot_window_custom_function


fabienrohrer:
👍

omichel:
:eyes:

omichel:
Looking at https://www.cyberbotics.com/doc/reference/robot?version=enhancement-underscores-in-title#wb_robot_window_custom_function it doesn't seem the problem is fixed: the underscores disappeared in `wb_robot_wwi_receive_text`.

fabienrohrer:
Oops you're right, it's still another issue, undescores should be protected in hyperlinks.

fabienrohrer:
👍

tests/test_paragraphs.py
```diff
@@ -46,10 +47,24 @@ def setUp(self):
...
+                linkName = m.group(1).strip()
+                # linkUrl = m.group(2).strip()
+                self.assertTrue(re.search(r'[^\\]_', linkName) is None, msg='Hyperlink "%s" contains unprotected underscore(s).' % m.group(0))
+
+    def test_hyperlinks_do_not_contain_prohibited_characters(self):
+        """Test that hyperlinks are not containing prohibited characters (such as '<')."""
+        for p in self.paragraphs:
+            for m in re.finditer(self.hyperlinkRE, p):
+                linkName = m.group(1).strip()
+                # linkUrl = m.group(2).strip()
```
omichel: Comment out code.

tests/test_paragraphs.py
```diff
@@ -46,10 +47,24 @@ def setUp(self):
...
                 linkName = m.group(1).strip()
                 # linkUrl = m.group(2).strip()
                 self.assertFalse('`' in linkName, msg='Hyperlink "%s" contains code.' % m.group(0))
+
+    def test_underscores_in_hyperlinks_are_protected(self):
+        """Test that every underscores appearing in hyperlinks are protected."""
+        for p in self.paragraphs:
+            for m in re.finditer(self.hyperlinkRE, p):
+                linkName = m.group(1).strip()
+                # linkUrl = m.group(2).strip()
```
omichel: Commented out code.

fabienrohrer:
In this case, I think the best is to reformulate this kind of sentences to not arrive in this situation.

omichel:
Or to allow hyperlinks in `fixed` mode and display them as hyperlinks.

fabienrohrer:
Yes. In this case, I have to create a new representation for hyperlink in fixed mode, and I fear it will impact the doc readability (My initial intention was precisely to avoid this). I'm trying to do so.

omichel:
I don't believe it will impact the code readability. I believe the CSS should only change the color of the hyperlink text similarly to the plain text.

fabienrohrer:
Ok, gray colors can be reserved for hyperlinks...

fabienrohrer:
Here is the result, it seems indeed good:

doc

I refactored the hyperlink test accordingly and found several other issues, notably in the matlab API.

fabienrohrer:
👍

omichel:
The appearance now looks very good to me.

But on this page: https://www.cyberbotics.com/doc/reference/robot?version=enhancement-underscores-in-title#wb_robot_window_custom_function if I click on the `wb_robot_wwi_receive_text()`, I don't land at the expected anchor... It seems broken.

fabienrohrer:
I can indeed reproduce, but even on master. I'm looking at this in another PR.

omichel:
I believe functions are referred to in an non-consistent way: sometimes we append the `()` and sometimes not. We should define a rule for that and remain consistent.

fabienrohrer:
I agree, this is currently not consistent, but it is also out of the scope of this PR.

fabienrohrer:
Do you want that I fix this here?

omichel:
I believe we should never add the `()` as it not a common usage in docs and could be misleading: one could think the function receives no argument.

fabienrohrer:
Ok for me, but this will lead to a big diff. Could I do this in another PR?

omichel:
Yes, but only for the code you modified, not the existing one. The existing one can be fixed in a separate PR. I will open an issue about it.

fabienrohrer:
Ok, I'm removing these addition only.

fabienrohrer:
At term, it would be also good if all these function references were hyperlinks.

fabienrohrer:
Ok, I only removed the ones I introduced.

👍

PR #573, opened 2018-02-01
Attempt to increase doc readability

fabienrohrer:
Aim: add a max-width for the content display, in order to increase readability.

Example: https://www.cyberbotics.com/doc/guide/webots-devices?version=enhancement-doc-readability

### DIFF
--- a/css/webots-doc.css
+++ b/css/webots-doc.css
@@ -47,10 +47,10 @@
   margin: 0 auto;
   background-color: white;
   max-width: calc(1000px - 4em);
-  padding: 40px;
 }
 .webots-doc #view {
   padding: 10px 40px 60px 40px;
+  margin-bottom: 30px;
 }
 .webots-doc #title {
   padding: 40px 40px 10px 40px;
--- a/css/webots-doc.css
+++ b/css/webots-doc.css
@@ -49,8 +49,11 @@
   max-width: calc(1000px - 4em);
   padding: 40px;
 }
+.webots-doc #view {
+  padding: 10px 40px 60px 40px;
+}
 .webots-doc #title {
-  padding-bottom: 0px;
+  padding: 40px 40px 10px 40px;
 }
 .webots-doc #navigation,
 .webots-doc #menu {
--- a/css/webots-doc.css
+++ b/css/webots-doc.css
@@ -33,12 +33,10 @@
   top: 50%;
 }
 .webots-doc #center {
-  top: 0px;
+  background-color: #eee;
   left: 20%;
   width: 80%;
   position: relative;
-  padding-bottom: 0;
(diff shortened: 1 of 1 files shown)


fabienrohrer:
@all I think this modification is slightly increasing the doc readability (in full screen - otherwise it is the same). Having marges at left and right is reducing the extremely long sentences. What do you think?

tn12787:
I think this is much better! Make sure the max-width is relative to the container width for responsivity reasons, however. Doing this also implies you should add a min-width so the doc isn't unreadable on small screens.

omichel:
It looks good to me.

I am wondering if the left and right margin (when going above the max-width) shouldn't be displayed as a grey background. That could look even nicer.

stefaniapedrazzi:
I think both have advantages and drawbacks.
About the readability, it is maybe slightly better.
But personally I don't like very much these big white margins that are now displayed on big window size and we loose space and we have to scroll more.
in any case any of the two version is good for me.

But in this page, the final note is not completely visible:
https://www.cyberbotics.com/doc/guide/controller-programming?version=enhancement-doc-readability

fabienrohrer:
> I think this is much better! Make sure the max-width is relative to the container width for responsivity reasons, however. Doing this also implies you should add a min-width so the doc isn't unreadable on small screens.

I will give a try, but I'm not sure it's doable simply. Adding a min-width is 


> I am wondering if the left and right margin (when going above the max-width) shouldn't be displayed as a grey background. That could look even nicer.

I can also give a try.

> I think both have advantages and drawbacks.
About the readability, it is maybe slightly better.
But personally I don't like very much these big white margins that are now displayed on big window size and we loose space and we have to scroll more.
in any case any of the two version is good for me.

Ok. Readability is obviously subjective, I'm also not sure yet at 100% it's better.

> But in this page, the final note is not completely visible:
https://www.cyberbotics.com/doc/guide/controller-programming?version=enhancement-doc-readability

This should be fixed in my last commit.

stefaniapedrazzi:
> This should be fixed in my last commit.

Yes, it is fixed now. Maybe it took some time to be available.

fabienrohrer:
rawgit always takes few minutes to be refreshed :-( It's annoying.

omichel:
We could implement our own version of rawgit... But I am not sure if the lag is in rawgit or in raw.githubusercontent.com...

fabienrohrer:
The doc layout is not logically split, causing several issues. I will try to fix this here.

fabienrohrer:
I need to think about this. I'm stopping this for now.

omichel:
Shall we close this one?

fabienrohrer:
Yes, thank you.

PR #575, opened 2018-02-01
Fix incorrect blog font

### LINKED ISSUE #574: The main blog font no longer loads
This is due to the new rawgit system trying to load the blog font from git, when it's actually on the FTP.

It's originally taken from the Google font cdn, so reverting back to loading it from the cdn should fix this problem and reduce problems like this in the future.

tn12787:
fixes #574.

This fixes the blog font by including it in the webots-doc submodule directly so it can be loaded from rawgit instead of the ftp server.

Will need a dual PR to update the packaging files.


tn12787:
👍

css/webots-doc.css
```diff
@@ -283,9 +283,16 @@
   margin: 0px;
 }
 
+@font-face {
+  font-family: "Roboto";
+  font-style: normal;
+  font-weight: 100;
+  src: local("fonts/Roboto"), local("fonts/Roboto"), url("fonts/Roboto.ttf") format("truetype");
```
omichel: `local("fonts/Roboto")` seems redundant, is that correct?

tn12787: You're right, it seems this can be removed. I will make a new PR to do this now.

css/main.css
```diff
@@ -16,5 +16,5 @@ body {
   font-family: "Raleway";
   font-style: normal;
   font-weight: 100;
-  src: local("Raleway Thin"), local("Raleway-Thin"), url("Raleway-Thin.ttf") format("truetype");
+  src: local("fonts/Raleway Thin"), local("fonts/Raleway-Thin"), url("fonts/Raleway-Thin.ttf") format("truetype");
```
omichel: Similarly, we should probably remove of these two local references.

tn12787: I think this is on purpose, font-face can be a pain with hyphens depending on browser/OS.

PR #583, opened 2018-02-06
Fix non-breaking string

fabienrohrer:
I applied an exhaustive search on `\*\*,\*\*` and this occurs only in `camera.md`.

### DIFF
--- a/reference/camera.md
+++ b/reference/camera.md
@@ -532,7 +532,7 @@ file extension).
 
 **Name**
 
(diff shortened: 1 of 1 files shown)


fabienrohrer:
Before: https://www.cyberbotics.com/doc/reference/camera#wb_camera_has_recognition
After: https://www.cyberbotics.com/doc/reference/camera?version=fix-individible-strings#wb_camera_has_recognition

👍

PR #592, opened 2018-02-07
Fix RawGit Not Refreshing Resources on Master

### LINKED ISSUE #589: The Master Snapshot from the RawGit CDN will never refresh.
https://github.com/rgrove/rawgit/issues/41

We need instead of loading just the page from the master branch (which is cached one and NOT flushable as described in the link above) we need some kind of unique identifier to get the version of the file we actually want.

I propose adding the master branch's HEAD commit SHA in order to create the unique URL. This way changes merged to master will actually make the page update.

tn12787:
fixes #589.

This PR ensures that when a commit is added to `master` the RawGit CDN resources are refreshed, avoiding the CDN caching mechanism which caches a given URL once *permanently*.

### DIFF
--- a/doc.php
+++ b/doc.php
@@ -59,8 +59,7 @@
         'anchor':     window.location.hash.substring(1),
         'branch':     '$branch',
         'repository': '$repository',
-        'url':        'https://raw.githubusercontent.com/$repository/webots-doc/',
-        'rawgiturl':  '$rawgiturl'
+        'url':        'https://raw.githubusercontent.com/$repository/webots-doc/'
       }
       console.log('Setup: ' + JSON.stringify(setup));
     </script>
--- a/doc.php
+++ b/doc.php
@@ -36,21 +36,21 @@
   if (!isset($repository))
     $repository = 'omichel';
 
-  # get HEAD commit SHA, to ensure that when master is updated the latest version is cached by the CDN
-  ini_set('user_agent', 'omichel'); # every GitHub request needs a valid user agent header
-  $githubHead = file_get_contents("https://api.github.com/repos/omichel/webots-doc/git/refs/heads/master");
-  // failed request / github is down
-  if ($githubHead === FALSE)
-    $sha = "master";
-  else {
-    $githubPhp = json_decode($githubHead);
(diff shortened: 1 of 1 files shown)


tn12787:
👍 (already in production).

n.b. this patch is working if this url has the new css: https://www.cyberbotics.com/doc/reference/radar

doc.php
```diff
@@ -47,7 +59,8 @@
         'anchor':     window.location.hash.substring(1),
         'branch':     '$branch',
         'repository': '$repository',
-        'url':        'https://raw.githubusercontent.com/$repository/webots-doc/'
+        'url':        'https://raw.githubusercontent.com/$repository/webots-doc/',
+        'rawgiturl':  '$rawgiturl'
```
fabienrohrer: If I'm correct, this seems to be unused.

tn12787: it is unused, but I figured it would be useful for debugging purposes, no?

fabienrohrer: I would say no: when we would like to debug, we have to introduce new variables etc., but introducing them in prevision to debug later is complexifying the code.

tn12787:
👍

doc.php
```diff
@@ -35,8 +35,20 @@
   }
   if (!isset($repository))
     $repository = 'omichel';
+
+  # get HEAD commit SHA, to ensure that when master is updated the latest version is cached by the CDN
```
fabienrohrer: All this sequence is only used when `$branch === ''`, it would seem more logical to me to move it in the test below.

stefaniapedrazzi:
Just note that uploading files on the FTP and put the in production before review is a bad habit.

tn12787:
Sorry, it's an ugly habit, yes.

tn12787:
GitHub has strict rate limiting, so I've added a fallback to use the dev version of rawgit at the **absolute worst**, in case GitHub refuses the request for the head SHA.

In general though, the same CDN-based mechanism will be used first. This way we can always load the current master branch snapshot, but with the consequence that it could be a little slower to load only if GitHub refuses our API request.

tn12787:
👍

PR #594, opened 2018-02-07
Fix cariage returns after etc. statements

fabienrohrer:
Some missing carriage returns after `etc.` where remaining.


fabienrohrer:
👍

PR #604, opened 2018-02-08
Replace double-quoted by single-quoted strings in JS

fabienrohrer:
Simply replaced double-quoted by single-quoted strings in JS files to match the ESLint specs.

There shouldn't be any impact on the behavior.

Test link: https://www.cyberbotics.com/doc/guide/index?version=enhancement-js-strings

### DIFF
--- a/js/showdown-extensions.js
+++ b/js/showdown-extensions.js
@@ -6,7 +6,7 @@ function wbSlugify(obj) {
   } else if (obj instanceof HTMLElement) {
     text = obj.textContent;
   } else {
-    console.error("wbSlugify: Unsupported input");
+    console.error('wbSlugify: Unsupported input');
   }
   return text.
     trim().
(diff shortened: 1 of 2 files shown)


fabienrohrer:
👍

stefaniapedrazzi:
👀

js/showdown-extensions.js
```diff
@@ -21,18 +21,18 @@ function wbSlugify(obj) {
...
-showdown.extension("wbVariables", function() {
+// For example, the markdown string '{{ date.year }}' is replaced by '2016'
+showdown.extension('wbVariables', function() {
   // static variables to maintain
   // TODO: could be computed
   var vars = {
     webots : {
       version : {
-        major : "R2018a",
+        major : 'R2018a',
```
stefaniapedrazzi: Note that this are updated through a script and the script has to be updated as well otherwise it won't work: `src/bin/new_version.sh`

fabienrohrer:
Well spotted. I'm creating a dual PR and will merge them at the same time.

PR #605, opened 2018-02-08
Fix differential wheels note display

fabienrohrer:
Before: https://www.cyberbotics.com/doc/reference/differentialwheels
After: https://www.cyberbotics.com/doc/reference/differentialwheels?version=fix-differential-wheels-note-display


fabienrohrer:
👍

PR #608, opened 2018-02-08
Merge webots r2018a revision 1 to develop

fabienrohrer:
Fixed conflicts in showdown-extensions.js and brake.md


js/showdown-extensions.js
```diff
@@ -1,177 +1,181 @@
...
       date : {
         year : 2018
       }
-    };
-    // compute debian package version format by removing initial 'R'
-    vars.webots.version.debian_package = vars.webots.version.package.substring(1);
+    },
+    date : {
+      year : 2018
+    }
```
DavidMansolino: Sorry to see this only now (because it probably comes from your previous merge) but it seems there is a problem here, the date is defined twice.

fabienrohrer: ach, well spotted: it's fixed here: https://github.com/omichel/webots-doc/pull/609

fabienrohrer:
👍

Issue #612, opened 2018-02-09
Strange rendering of quoted file path


stefaniapedrazzi:
Some parts of `"WEBOTS_HOME\projects\robots\aldebaran\controllers\robocupstriker\robocupstriker.sln"`are written in italic:
https://www.cyberbotics.com/doc/guide/using-visual-cpp-with-webots

DavidMansolino:
Fixed in #613

PR #613, opened 2018-02-09
Fixed underscore in using-visual-cpp-with-webots

DavidMansolino:
Fix issue  #612

https://www.cyberbotics.com/doc/guide/using-visual-cpp-with-webots?version=fix-underscore


DavidMansolino:
👍

PR #618, opened 2018-02-13
Pioneer example description

remidhum:
Adding screenshots and descriptions for the two Pioneer robots' world examples.


guide/using-the-pioneer-3-at-robot.md
```diff
@@ -25,6 +25,13 @@ More information on the specifications and optional devices is available on Adep
 %end
 
 The standard model of the Pioneer 3-AT is provided in the "pioneer3AT.wbt" file which is located in the "WEBOTS\_HOME/projects/robots/adept/pioneer3/worlds" directory of the Webots distribution.
+
+%figure "Pioneer 3-AT world example"
+
+![pioneer3at_example.png](images/pioneer3at_example.png)
+
+%end
+
```
omichel: A figure should always be referred to from the text. So you may add some "(see figure xxx)" in the above text or the like.

fabienrohrer:
I think we should not show the Webots interface in the samples, because it will be difficult for us to maintain. Could you crop this screenshot to see only the 3D view?

guide/using-the-pioneer-3-at-robot.md
```diff
@@ -45,3 +44,13 @@ The sonar sensors are numbered according to [this figure](#sonar-sensors-positio
 %end
 
 The angle between two consecutive sensor directions is 20 degrees except for the four side sensors (so0, so7, so8 and so15) for which the angle is 40 degrees.
+
+### Pioneer 3-AT obstacle avoidance
+
+The "pioneer3at.wbt" world file is a simulation example of a simple obstacle avoidance behaviour based on the use of a SICK LIDAR (see the "obstacle\_avoidance\_with\_lidar.c" controller file in the "WEBOTS\_HOME/projects/robots/adept/pioneer3/controller" directory).
```
omichel: We use American English: behaviour &rarr; behavior

tn12787:
Especially because the Webots night will be the default theme 😉

guide/using-the-pioneer-3-at-robot.md
```diff
@@ -45,3 +44,13 @@ The sonar sensors are numbered according to [this figure](#sonar-sensors-positio
...
+
+%figure "Pioneer 3-AT world example"
+
+![pioneer3at_example.png](images/pioneer3at_example.png)
+
+%end
```
omichel: A figure should always be referred to in the text.

remidhum:
👍

fabienrohrer:
I would change a little bit the title levels (to match the "Webots Samples Application" chapter), and use the new illustrated paragraphs instead of figures for the examples:

```md
### Samples

#### pioneer3at.wbt

![pioneer3at_avoidance.png](images/pioneer3at_avoidance.png) The "pioneer3at.wbt" world file is a...

#### other_simulation.wbt

![...](...) Blah blah.

```

fabienrohrer:
I think it would be better if the image file names match with the wbt files (if possible): i.e. Rename `pioneer3dx_avoidance.png` =>`pioneer3dx_collision_avoidance.png`

fabienrohrer:
I think it would be good to be as exhaustive as possible, and `pioneer3dx_matlab.wbt` is not documented here.

remidhum:
Should I include something extra for the particular case of the `pioneer3dx_matlab.wbt` as the screenshot is the exact same one as the `pioneer3dx.wbt`?

fabienrohrer:
> Should I include something extra for the particular case of the pioneer3dx_matlab.wbt as the screenshot is the exact same one as the pioneer3dx.wbt?

I like the rule to have one screenshot for each sample, even if it's the same screenshot.

remidhum:
👍

omichel:
It would be nice to post test links along with the 👍 when such a PR is ready for review, like https://www.cyberbotics.com/doc/guide/using-the-pioneer-3-dx-robot?version=enhancement-pioneer-example-description and https://www.cyberbotics.com/doc/guide/using-the-pioneer-3-at-robot?version=enhancement-pioneer-example-description

omichel:
This table doesn't look very readable:
image
We should maybe have one numerical value per line to make it look better.

omichel:
I believe we should define a precise font family, font weight and font size to use on all the images where we need to write some text. In this PR, I already see two different font sizes on two different images...

remidhum:
I'll redo all the images with Sans at 30px after rescale (width = 700px).

omichel:
OK, I assume "Sans" refer to the font named like that in Gimp.

remidhum:
👍 
https://www.cyberbotics.com/doc/guide/using-the-pioneer-3-at-robot?version=enhancement-pioneer-example-description

omichel:
Sorry, but I believe this font is too big. That doesn't look nice:
image
Can't we improve that?

remidhum:
The problem is that the text would be a bit too small for the other images...
I can put the names on the top and bottom, thus increasing the relative size of the robot in this picture wich would make the text seem smaller.

remidhum:
After a discussion with @omichel, the possibility of setting the size of the different images has arizen. Indeed, since the font of the images does not match the one from the text and I feel like some robots are too big, the image sizes could be manually controlled. 

capture
capture2

This can be done using:
``

for example : 
``
becomes
``

This would still enforce the image standard from previously but allow manual control for aesthetics.

fabienrohrer:
You can force image sizes using this MD notation:

```Markdown
![](images/mypng.png =26x26)
![](images/mypng.png =200x)
```

remidhum:
**Image standard:**
Width: 700px (here, most robots were scaled down by a factor of 2 in the previous 700px canvas, to compensate for the scale, 100px were removed in height)
Shadows: 60% transparency (ground)
Text: Sans (Default on Gimp) @ 15px in bold

remidhum:
👍 
https://www.cyberbotics.com/doc/guide/using-the-pioneer-3-at-robot?version=enhancement-pioneer-example-description

fabienrohrer:
=> If the images are cropped (Gimp auto crop feature), then the result is identical.

remidhum:
But that would result in storing bigger images than neccessary no? (which is something @omichel wanted to avoid)

fabienrohrer:
No, the CSS is only applying max-width when the required, for example if the window size is too small. If it's not the case like here, the image is shown at 100%, there is no information loosed.

Note that I'm saying to crop their empty spaces, not to scale them.

I about to commit this.

PR #623, opened 2018-02-13
More Modern Sidebar

tn12787:
Preview: https://www.cyberbotics.com/doc/reference/positionsensor?version=enhancement-better-sidebar


tn12787:
👍 https://www.cyberbotics.com/doc/reference/positionsensor?version=enhancement-better-sidebar

css/webots-doc.css
```diff
@@ -17,6 +17,9 @@
   src: local("fonts/Roboto-Mono"), url("fonts/Roboto-Mono.ttf") format("truetype");
 }
 
+b, strong {
```
fabienrohrer: Is this still required?

tn12787: It's better practice, but it's technically not needed, so I removed it.

PR #633, opened 2018-02-15
Better robot menu

remidhum:
Improved robot visibility in menu by using images.

Preview: https://www.cyberbotics.com/doc/guide/robots?version=emhancement-better-robot-menu

PR #636, opened 2018-02-16
Creating the Aibo ERS7 doc page

remidhum:
In an effort to continue the Robot documentation standardization, the Aibo ERS7 is being moved from the "Real robots" and having its own dedicated page based on the ones from the Pioneer robots.

Preview: https://www.cyberbotics.com/doc/guide/using-the-aibo-ers7-robot?version=enhancement-aibo-ers7

This branch should be merged after #633.


remidhum:
👍

PR #637, opened 2018-02-16
Fix doc "top" height

tn12787:
This PR prevents this ugly 2px line on top of the doc: 
ugly pixels

This isn't visibly locally but needs updating in production.


tn12787:
👍

PR #640, opened 2018-02-16
Fix accordion symbols

tn12787:
Fix unwanted accordion symbols on active menu items in the sidebar in the online doc.


tn12787:
https://www.cyberbotics.com/doc/reference/motion-functions?version=fix-accordion-symbols

👍

PR #651, opened 2018-02-21
Add doc about the e-puck slowdown issue resolution

fabienrohrer:
Ditto


fabienrohrer:
I have a doubt it's our responsibility to document this, but this small note doesn't hurt too.

👍

Issue #657, opened 2018-02-22
Use of Title Case


tn12787:
In English (British or American), words in titles are all capitalised (with the exception of some connectives such as and, of, or).

It would be good to ensure this is the case in all titles.

fabienrohrer:
.. and add a test for this :innocent:

PR #659, opened 2018-02-22
Correct title capitalization

fabienrohrer:
Attempt to fix https://github.com/omichel/webots-doc/issues/657

- [x] Fix/uniformize titles
- [x] Add a test

Preview: https://www.cyberbotics.com/doc/guide/index?version=fix-capitalized-title


fabienrohrer:
👍

fabienrohrer:
👍

reference/callback-functions.md
```diff
@@ -88,24 +88,24 @@ Since Webots 7.2.0, a multi-threaded version of ODE is used.
 Therefore, this function may be called from different threads.
 You should ensure it is re-entrant and that every ODE function call modifying the ODE world is protected by mutexes as explained earlier.
 
-### void webots\_physics\_step()
+### "void webots\_physics\_step()"
```
omichel: Didn't we decide to not use the parenthesis for referring to a function?

guide/controller-plugin.md
```diff
@@ -195,7 +195,7 @@ In the motion editor different fonts and colors are used to identify the status
 
 %end
 
-### Remote-control Plugin
+### Remote-Control Plugin
```
omichel: Inconsistency: sometimes we write "remote control" (without dash) and sometimes, it's "remote-control" (with dash). We should be more consistent about this.

tn12787: To be specific: "Remote Control" has no hyphen. "Remote-controlled" (using a passive form of control) needs a hyphen.

reference/callback-functions.md
fabienrohrer: In this case it's not obvious, because later in this file, we have similar titles but with arguments.

guide/controller-plugin.md
fabienrohrer: I propose to fix this in another PR, because it goes quite far (I just checked and there are 130 references to remote control in various states).

fabienrohrer:
👍

PR #624, opened 2018-02-13
WREN

luc4s:
This PR is a base branch for modifications specifically related to WREN, which shall be merged when WREN is merged into develop.

PR #670, opened 2018-03-02
Feature Point Set

DavidMansolino:
Linked to omichel/webots#6882


DavidMansolino:
👍   (I will merge only after #624)

reference/pointset.md
```diff
@@ -0,0 +1,22 @@
...
+### Description
+
+The [PointSet](#pointset) node represents a 3D geometry formed by a set of 3D points specified in the `coord` field.
+[PointSet](#pointset) nodes are not lit, are not texture-mapped and they do not cast or receive shadows.
+[PointSet](#pointset) nodes cannot be used for collision detection (boundingObject).
+
+### Field Summary
+
+The `color` field optionally contains a [Color](color.md) node which defines the color of each point.
+If the `color` field is not NULL, the [Color](color.md) node should have the same number of items than the [Coordinate](coordinate.md) node in the `coord` field.
```
omichel: than &rarr; as

reference/pointset.md
```diff
@@ -0,0 +1,22 @@
...
+If the `color` field is NULL, the `emissiveColor` of the [Material](material.md) will be used to define the color of the points. 
+
+The `coord` field specifies the list of 3D points and contains a [Coordinate](coordinate.md) node.
```
omichel: I would say: "The `coord` field contains a [Coordinate](coordinate.md) node that specifies the list of 3D points."

DavidMansolino:
> On the node hierarchy image, we should try to better respect the alphabetical order. It seems it was originally respected, but when adding new nodes, they are randomly added or added at the end of the list, whereas they should have been added in the alphabetical order.

I will take advantage of this PR to fix this (note that I did added the `PointSet` at the end of the list because it is indeed the last one (alphabetically speaking)).

reference/pointset.md
```diff
@@ -0,0 +1,22 @@
...
+```
+PointSet {
+  SFNode  color   NULL  # {Color, PROTO}
+  SFNode  coord   NULL  # {Coordinate, PROTO}
+}
+```
+
+### Description
+
+The [PointSet](#pointset) node represents a 3D geometry formed by a set of 3D points specified in the `coord` field.
```
tn12787: I would remove the "represents a 3d geometry formed by" - it's not really correct and "a geometry" isn't right either.

DavidMansolino:
👍

DavidMansolino:
I am starting to have something functional with tooltips and link on nodes (colors are only for testing, final colors will change to suit better the one of the doc):
test

The flowchart of https://mermaidjs.github.io/flowchart.html seems very powerful as we can change the color/style of almost every element (and we can even do this using a CSS file). And it does have a nice visualization in Atom 😉 
The only drawback I see is that we don't have full control on where are placed the elements and on the path of the array (especially when the graph becomes long). But I don't think this is a big problem as it will probably be clearer to split this chart in several thematic smaller flowcharts (e.f. Group, Geometry, etc.), on the same page of course.

DavidMansolino:
I am having issues to add the 'Device' layer and keep the main graph readable (not all the devices are present and it quickly become a mess):
chart1
I can instead add a group which is easily readable but not very clear in my opinion:
flow2
Or I can add the 'Device' between the 'Solid' and the actual devices, but it is not 100% correct for the JointDevice (even if in the documentation we said that jointDevices inherits from Device which inherits from Solid):
flow3

DavidMansolino:
Remains to align + homogenize width:
flow6

DavidMansolino:
With WRML indication:
flow1

DavidMansolino:
Am have having a lot of issues with links and the size.
I have found a workaround that let the links works but I can't find any way to fix the size of the graph so that it is not truncated:
capture
The configuration about the width restriction in the mermaid API doesn't seems to do anything.
And for some reason the height of the SVG is always 150px (whatever the size of it's content is), I can override this value from the CSS but only to a fix value in pixel (auto or percentage gives a height of 150px too).

DavidMansolino:
Current result integrated in the doc:
capture

I will now port this to master.

omichel:
It looks like many node names are cut (last letter or so).

DavidMansolino:
> It looks like many node names are cut (last letter or so).

Yes indeed I saw this too. It seems it has something to do with the links as this happens only for labels which are links.

DavidMansolino:
👍

DavidMansolino:
👍    (need a re-approval because of merge)

PR #673, opened 2018-03-06
Use querySelector API where possible

tn12787:
ditto.

PR #676, opened 2018-03-08
Fix Code Highlighting

tn12787:
This PR fixes an issue with code highlighting from #673, simplifying the selector process for getting code blocks.


tn12787:
preview: https://www.cyberbotics.com/doc/reference/brake?version=fix-broken-code-highlighting

tn12787:
👍

fabienrohrer:
👀

fabienrohrer:
Could you explain why you changed "c++" to "cpp". A priori it seems less readable to me.

tn12787:
`.c++` is not a valid CSS selector (you can't use characters like `+`) whereas they can be used in HTML5 attributes. If we use `getElementsByClassName` we can use a selector like `.c++` but using query selector this does not work so we must use `.cpp` instead.

PR #678, opened 2018-03-08
Remove Reference to "--enable-ogre-warnings"

tn12787:
dual with https://github.com/omichel/webots/pull/6934


tn12787:
👍

DavidMansolino:
Note that the 'WREN' PR has not yet been reviewed, therefore if we don't want to double-review the content of this PR we should review the other PR before merging this one.

PR #689, opened 2018-03-09
Fix Node Hierarchy

DavidMansolino:
It was wrong to say that `Device` derived from `Solid`, it is not the case and could lead the user to think that any device have all the solid fields.


DavidMansolino:
👍

PR #690, opened 2018-03-12
Responsive robot list

fabienrohrer:
The robot list doesn't support well small windows size because of the image max-width and of the table header wrap.

Without patch:

https://www.cyberbotics.com/doc/guide/robots
without patch

With patch:

https://www.cyberbotics.com/doc/guide/robots?version=enhancement-responsive-robot-list
with patch


fabienrohrer:
👍

PR #691, opened 2018-03-12
Fix the contribute banner on macOS when embedded in Webots

fabienrohrer:
The main issues are:
1. -webkit extension should be used for the translation rules.
2. left is not computed the same in webkit (we expect "inherit")

Notes:
- "position: -webkit-sticky" is not used (overhiden just bellow)
- "margin" is useless

With patch:

with patch 17 02 34

Without patch:

without patch 17 02 34


fabienrohrer:
👍

fabienrohrer:
👍  (need a reapproval after merging master)

PR #711, opened 2018-03-16
Fixed crash with offline documentation

omichel:
Fixes https://github.com/omichel/webots/issues/6762

We should not use non-ASCII characters, e.g., UTF-8 in our MD files as it seems to crash QtWebKit (at least on Windows). I tried to fix it by forcing UTF-8 encoding in Webots, but that didn't work. I also tried to force UTF-8 encoding in `viewer.js` but that didn't work either. Therefore, I suspect a bug in QtWebKit about it. This patch works around this bug.


omichel:
Strangely, other unicode characters don't crash webots, such as accented characters and the '®' character.

PR #717, opened 2018-03-19
 Add missing OSM importer dependency on Ubuntu 14.04

stefaniapedrazzi:
On Ubuntu 14.04 I had to explicitly install the `configparser` python modules.
This seems related to the changes on https://github.com/omichel/webots/pull/6709.

I still have to test on a clean Ubuntu 16.04 environment to check if it has to be manually installed as well.


stefaniapedrazzi:
The same dependency is also now required on Ubuntu 16.04.

It should probably also be tested on macOS and Windows.

:+1:

stefaniapedrazzi:
:+1:

fabienrohrer:
I agree with the David solution. It's difficult to me to test this on mac, because I'm far to have a clean install.

DavidMansolino:
I can now confirm this module has to be installed on macOS too.

fabienrohrer:
👍

PR #737, opened 2018-03-26
Fix anchor test and related issues.

fabienrohrer:
The test about anchors referring to an anchor in the current file was failing silently.
Once fixed it reveals several anchor issues, which are fixed here.


fabienrohrer:
The test fails now correctly when required: https://travis-ci.org/omichel/webots-doc/builds/358387090?utm_source=github_status&utm_medium=notification

fabienrohrer:
👍

PR #738, opened 2018-03-27
Automatic index

fabienrohrer:
Preview: https://www.cyberbotics.com/doc/guide/robotis-op2?version=feature-automatic-index

PR #739, opened 2018-03-27
Fix index which cannot take into consideration a double level jump.

fabienrohrer:
Hotfix: #738


fabienrohrer:
Normally, this patch doesn't affect the current state, because this case is rare, but it's visible in the David PR about objects.

fabienrohrer:
👍

DavidMansolino:
👀

PR #725, opened 2018-03-21
Feature PROTO doc Generator

DavidMansolino:
Preview:
https://www.cyberbotics.com/doc/guide/objects?version=feature-proto-doc-generator
https://www.cyberbotics.com/doc/guide/object-road?version=feature-proto-doc-generator
https://www.cyberbotics.com/doc/guide/object-factory?version=feature-proto-doc-generator

https://www.cyberbotics.com/doc/automobile/menu?version=feature-proto-doc-generator

Dual PR: omichel/webots#7061

Any preliminary comments are welcome (not on the content of course as this should be improved in the PROTO files directly).

For now I am using an inclusive list of PROTO to test, but for the final version I plan to transform this into an exclusive list instead in order to not forget any future PROTO files.

Tasks:
  - [x]  Skip hidden and deprecated PROTOs
  - [x] Fix menu item order.
  - [x] Fix empty image.
  - [x] Fix image transparent.
  - [x] Fix detection of description end.
  - [x] Escape problematic caracters.
  - [x] Remove automobile objects.

PR #742, opened 2018-03-28
Document the robot PROTO nodes and their fields

fabienrohrer:
Extract robot PROTO fields and add them in the documentation.

Constraint: be coherent with https://github.com/omichel/webots-doc/pull/725

Preview: https://www.cyberbotics.com/doc/guide/robots?version=enhancement-document-robot-fields


fabienrohrer:
@DavidMansolino @omichel Is it ok to scale with this?

omichel:
Yes, that's fine for me (except that the qrio.wbt is not related with the sojourner).

fabienrohrer:
> except that the qrio.wbt is not related with the sojourner

Sorry but I don't understand this comment.

fabienrohrer:
oh I see, this title is wrong: https://www.cyberbotics.com/doc/guide/sojourner#qrio-wbt

DavidMansolino:
That's fine for me, it is now coherent with #725.

omichel:
On the preview page, there is something wrong:
image

DavidMansolino:
Note that I just write `Sojourner Field Summary` (without repeating the 'PROTO' keyword).

fabienrohrer:
👍

fabienrohrer:
👎  (tests are failing)

fabienrohrer:
Now they are happy!

fabienrohrer:
👍

omichel:
The translation, rotation and name fields are not described for object PROTO nodes. We should not add them in the robot PROTO nodes. Also, I am wondering whether it is good or not to repeat all the Robot fields (inherited from Robot). That is very redundant...

fabienrohrer:
I think that the robot pages are an entree point in the documentation for new users who are not aware of the Robot node. It's clearly redundant, but not documenting them is also not precise.

omichel:
This is not consistent with #725 where translation, rotation and name are omitted.

fabienrohrer:
Sure.

A solution could be to add a "Derived from Robot node" note or so, but in this case it presume the user should guess there is a link.

DavidMansolino:
Note that in #725 a `Derived from XX.` is indeed added automatically at the end of the description.

omichel:
Yes, that seems to be a good solution.

fabienrohrer:
I'm still sketpical. I think that if a user search the fields in the documentation, it's precisely because he don't know from where they come from. It's only obvious for us. Moreover I don't where to stop with this: what about `controller`, `synchronisation` and even `camera_width`?

fabienrohrer:
Just discussed with Olivier and David, and we will go on with the solution about documenting the derived node, and removing the fields used directly in the PROTO root node from the "field summary".

fabienrohrer:
👍

omichel:
That looks wrong:
image

DavidMansolino:
For the BB8:

> maxVelocity: Defines the RotationalMotor fields. Note: the real robot moves at 4.9 [mph].

DavidMansolino:
Same kind of issue of the Khepera IV

fabienrohrer:
👍

PR #741, opened 2018-03-28
Enhancement of the index location and look

fabienrohrer:
- [x] Change the index rules to be created just before the second title.
- [x] Improve the index look.

Some previews:

- https://www.cyberbotics.com/doc/guide/cpp-java-python?version=enhancement-index
- https://www.cyberbotics.com/doc/guide/robotis-op2?version=enhancement-index
- https://www.cyberbotics.com/doc/reference/robot?version=enhancement-index

PR #745, opened 2018-03-28
Fix Chart Font Size.

DavidMansolino:
The size of the font was modified in #741, but in the case of charts it is breaking the layout:
  - https://www.cyberbotics.com/doc/guide/tutorial-2-modification-of-the-environment-20-minutes#def-use-mechanism-applied-on-the-shape-node-of-a-solid

Fixed on this branch:
 - https://www.cyberbotics.com/doc/guide/tutorial-2-modification-of-the-environment-20-minutes?version=fix-chart-font-size#def-use-mechanism-applied-on-the-shape-node-of-a-solid


DavidMansolino:
👍

PR #762, opened 2018-04-05
Fix DistanceSensor Documentation

### LINKED ISSUE #757: Lookup Table documented only for DistanceSensor
The lookup table model and how it works is described in detail only for the distance sensor:
https://www.cyberbotics.com/doc/reference/distancesensor#field-summary

A lot of other sensors are using lookup tables, we should at least add a link to the description of the lookup table in the distance sensors page.

DavidMansolino:
The documentation was not looking correct (see especially https://www.cyberbotics.com/doc/reference/distancesensor#sensor-response-versus-obstacle-distance)

This branch: https://www.cyberbotics.com/doc/reference/distancesensor?version=fix-distance-sensor-doc#lookup-table

I took advantage of this PR to fix #757


DavidMansolino:
👍

omichel:
:eyes:

PR #767, opened 2018-04-06
Regenerate Objects Doc for develop

DavidMansolino:
https://www.cyberbotics.com/doc/guide/object-buildings?version=regenerate-objects-doc
https://www.cyberbotics.com/doc/guide/object-street-furniture?version=regenerate-objects-doc
https://www.cyberbotics.com/doc/guide/object-garden?version=regenerate-objects-doc


omichel:
The SimpleBuilding also has field descriptions in the PROTO description part. That should be moved to the corresponding field description.

omichel:
In the PicketFence segment description, it would be more interesting to have the width of a segment rather than its height (length)...

DavidMansolino:
> The question mark on the random building icon is not very much visible. It would be more visible in white or on the red roof.

I prefer to keep it on the wall because it is the wall appearance that is random, but I will but it in white.

DavidMansolino:
> The SimpleBuilding also has field descriptions in the PROTO description part. That should be moved to the corresponding field description.

This last issue will be fixed once we have implemented omichel/webots#7158.


👍

PR #780, opened 2018-04-12
Dynamic load of dependencies

### LINKED ISSUE #778: Load CSS and JS dependencies dynamically from a file
viewer.js, the CSS and the MD files are loaded dynamically depending on the branch.

But it's not the case for the JS and CSS dependencies which are hard-coded:

https://github.com/omichel/webots-doc/blob/master/index.html#L6
https://github.com/omichel/webots-doc/blob/master/index.html#L39
https://github.com/omichel/webots-doc/blob/master/doc.php#L66

This is extremely annoying for testing in production modifications on these dependencies (update, addition, removal) slowing a lot the workflow.

I'm wondering if we could add them to a file and load them dynamically from webots.js.

fabienrohrer:
Fix #778

- [x] Create a `dependencies.txt` file listing all the CSS and JS deps (which are on the FTP)
- [x] Update doc.php to build on-the-fly the HTML tree using `dependencies.txt`
- [x] Update `local_exporter.py` to generate `index.html` (git-ignored) from `index.template.html` and `dependencies.txt``
    - [x] Remove `local_index.html`
- [x] Updated README.md accordingly


fabienrohrer:
It seems to work offline (I tested using `php -S localhost:8000 -t .`). Could I have an intermediate review and could I give a try online?

fabienrohrer:
👍

omichel:
That looks very good to me so far.

fabienrohrer:
Just discussed with Stefania. Even if it's annoying to touch the R2018a, it seems good to do it for this case, because it will increase our workflow and continuous integration directly.

fabienrohrer:
I'm giving a try in the FTP.

fabienrohrer:
I tested this PR quickly in production and it's working as expected.
However I can't push doc.php until we merge this on master (otherwise it breaks master).

👍

fabienrohrer:
I will merge this on Monday ;-)

omichel:
That sounds safer indeed 😀.

fabienrohrer:
👍

PR #783, opened 2018-04-16
Merge webots r2018a revision 2 to develop

fabienrohrer:
Fixed conflicts in index.template.html and local_exporter.py about dependencies (request_methods.js  has been updated on develop, entering in conflicts with the new dependency system).


fabienrohrer:
👍

PR #784, opened 2018-04-16
Fix local exporter to be called from another directory.

fabienrohrer:
This is required to be called smoothly from webots root Makefile.


fabienrohrer:
👍

PR #788, opened 2018-04-17
Improve Supervisor node_get_from_def description

stefaniapedrazzi:
Related to https://github.com/omichel/webots/pull/7225: better explain the behavior in case of internal PROTO nodes.


stefaniapedrazzi:
I'm still checking if this is really what we want.
For the moment it seems that it works only because the function is buggy.

stefaniapedrazzi:
The documentation is correct and describes what we expect, but the implementation doesn't match it.

PR #791, opened 2018-04-18
Fix Vertical Figure > Img Scrollbar

### LINKED ISSUE #787: Unneded vertical scrollbar for images
When opening the offline User Guide in Webots, I see a vertical scrollbar for the image.
This is not always visible but depends on the width of the documentation.

offline_guide

tn12787:
fix #787.

### DIFF
--- a/css/webots-doc.css
+++ b/css/webots-doc.css
@@ -332,7 +332,7 @@
   margin: 20px auto 20px auto;
 }
 .webots-doc figure > img {
-  overflow-x: initial;
+  overflow: visible;
   max-width: 80%;
 }
 .webots-doc figcaption {
--- a/css/webots-doc.css
+++ b/css/webots-doc.css
@@ -327,12 +327,11 @@
   vertical-align: middle;
 }
 .webots-doc figure {
-  overflow-x: auto;
+  overflow: visible;
   text-align: center;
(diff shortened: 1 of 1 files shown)


tn12787:
Since this bug is only reproducible embedded in Webots, I'm not posting a preview link.

tn12787:
:+1:

stefaniapedrazzi:
:eyes:

stefaniapedrazzi:
Did you check why this overflow property was introduced and that it still works correctly?

tn12787:
Yes, It was introduced by @fabienrohrer and I several months ago to remove some horizontal scrollbars on figures. It wasn't a very good solution as it only took care of the x direction and used "auto" so was up to the User Agent Stylesheet to do whatever it wanted. We always want figures to be completely visible and we have responsive widths set up such that we should never need a scrollbar (just like the unnecessary one here), so it is safe to disable scrolling on these figures.

stefaniapedrazzi:
Ok.
As far as I could see this was introduced by @DavidMansolino  in the charts feature PR.
But charts doesn't seem to be displayed correctly in any case.

DavidMansolino:
@tn12787 are you sure this doesn't break the chart?
It seems the horizontal scrollbar is not working on chart on the online doc now...

PR #792, opened 2018-04-18
Fix word wrapping on code blocks in embedded doc

tn12787:
Fix horizontal scrollbars on embdedded doc.

### DIFF
--- a/css/webots-doc.css
+++ b/css/webots-doc.css
@@ -240,7 +240,10 @@
   white-space: pre;
   word-wrap: break-word;
 }
-.webots-doc tr > td >  code, .webots-doc pre > code {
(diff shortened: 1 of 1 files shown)


tn12787:
I'm starting to really worry about boostrap applying endless rules on top of the doc css, it makes developing the actual doc css error prone (like this bug) and rather tedious. I think the doc css should apply a css reset   of some kind to undo bootstrap's influence.

fabienrohrer:
I agree with this observation, and open an issue there: https://github.com/omichel/webots-doc/issues/795

tn12787:
:+1:

DavidMansolino:
👀

PR #800, opened 2018-04-23
Robot component

fabienrohrer:
- [x] Robot component mechanism
    - [x] Add a showdown extension to be able to add X3D scene with robots.
    - [x] Add an interactive list of robot devices, showing
        - [x] Motor slider
        - [x] LED color
        - [x] Device center
    - [x] Add an interactive menu (reset viewpoint / motors) show/hide the menu
- [ ] Scale to all the robots
    - [x] Nao: https://www.cyberbotics.com/doc/guide/nao?version=feature-robot-component
    - [x] e-puck: https://www.cyberbotics.com/doc/guide/epuck?version=feature-robot-component
    - [ ] ...

Remaining bugs:

- [ ] Some solid IDs are badly referenced
- [ ] Some reset motor position are not working
- [ ] Infinite range are not working
- [ ] LED colors are hard-coded
- [ ] center.png icon is not scaled
- [ ] Proofread the


fabienrohrer:
As I feared, this component is not working well when embedded in Webots (macOS: crash in Ogre / linux&windows: OpenGL is not loaded / windows: crash on resize).

=> I will implement a fallback to the original image in this case. I think it will be much easier to fix this when using wren + QWebEngine.

fabienrohrer:
👍 I'm ready for an intermediate review before scaling to all the robots. For now, only the e-puck and the nao have been exported.

omichel:
I tried the NAO, but can't rotate the head or arms using the slider, but the fingers works...
Also, it's a pity that the range of the motors is not displayed (max and min positions).

fabienrohrer:
Oops, I recently broke this.

omichel:
I believe the tooltips are not very clear. What is a component?
I would replace "Reset the robot component." by "Reset Viewpoint" and "Show/Hide the device component." by "Show/Hide Device List".

Also the yellow/black target is often annoying: when I move the fingers, it hides the fingers and I cannot see very well their motion. I would make it more transparent or add an option to disable it.

DavidMansolino:
> the yellow/black target is often annoying: when I move the fingers, it hides the fingers and I cannot see very well their motion.

I had indeed the same issue with the LED of the e-puck when testing.

omichel:
The "Reset Viewpoint" icon doesn't look very good to me. I would rather use something like that: http://www.endlessicons.com/free-icons/refresh-icon/

omichel:
In the device list, I would re-order the devices. I would put first (on the top) the most interesting ones, that is the rotational motors, then the linear motors, then the LEDs and finally all the other devices, listed alphabetically.

fabienrohrer:
Thank you for the review, I agree with your comments. I'm doing a new iteration before scaling.

fabienrohrer:
👍 @omichel I think I adressed all your comments, a new intermediate review would be welcome.

omichel:
It looks much better this way! Thank you.
However, I have tested on Edge and there are problems: the buttons have no image and the sliders are too small and difficult to use:
image

omichel:
On Internet Explorer, it is even worst: the sliders are unusable...

fabienrohrer:
Ok, I will redo an iteration tomorrow on the Windows browser.

DavidMansolino:
I have another bug on edge, if I load the page, doesn't move the robot/motors at all, when pressing the 'reset' button the left arm moves to this position:
 
load             =>                  reset

fabienrohrer:
👍 It seems I fixed all the issues reported on the Windows browsers.

fabienrohrer:
New exported robots:

- https://www.cyberbotics.com/doc/guide/aibo-ers7?version=feature-robot-component
- https://www.cyberbotics.com/doc/guide/altino?version=feature-robot-component

New bugs:

- [ ] The aibo leds are buggy: https://www.cyberbotics.com/doc/guide/aibo-ers7?version=feature-robot-component (probably because the material is not applied only on the first LED child).
- [ ] Sometimes, the billboard icon is not visible.

DavidMansolino:
> https://www.cyberbotics.com/doc/guide/altino?version=feature-robot-component

Note that in the altino case I am not sure we are interested in the complete device list as most of them are internal to the car/driver library.

omichel:
> Note that in the altino case I am not sure we are interested in the complete device list as most of them are internal to the car/driver library.

I would keep this like that for now. It is more homogeneous with other robots, it shows the possibilities of the robot and using the car/driver library is not mandatory for using this robot. But maybe in a subsequent PR we could make a GUI that matches the car/driver library (e.g., and have for example both front wheels turning together in the same direction, all 4 four wheels rotating at the same speed, etc.).

fabienrohrer:
I think I agree with Olivier: it will be probably a good step to achieve when we will apply the robot component for the vehicles.

fabienrohrer:
I applied the robot component to all the robots, and the result is globally very good I think.

Here is a new bug list:

- [ ] Linear motors are not working (cf. youbot and pr2 fingers)
- [ ] IPR and youbot initial viewpoints are not good.

fabienrohrer:
👍

omichel:
On the Robotis OP2, LegLowerR/L motors seems to rotate the joint in the wrong direction.

fabienrohrer:
> On the Robotis OP2, LegLowerR/L motors seems to rotate the joint in the wrong direction.

Robot component is not taking into consideration the HingeJointParameters.position field. I will fix this.

fabienrohrer:
test me

fabienrohrer:
👍

omichel:
I am wondering why the iCub is missing. Any reason?

DavidMansolino:
> I am wondering why the iCub is missing. Any reason?

I asked @fabienrohrer yesterday and he explained me that it is because the iCub is not a PROTO node.

omichel:
I don't like very much the word "component" as it doesn't have a precise meaning... But I am not sure to find a better word for it. What about "panel"? Anything better?

fabienrohrer:
👍

guide/scenes/puma/puma.meta.json
```diff
@@ -0,0 +1,77 @@
...
+      "transformID": "n12870", 
+      "name": "joint_GripperR", 
+      "maxPosition": 0.01, 
+      "position": 0, 
+      "type": "LinearMotor", 
+      "minPosition": -0.004, 
+      "axis": "-1 0 0"
+    }
+  ]
+}
```
tn12787: missing newline at EOF (I think the same is true for `*.meta.json` in this PR)

fabienrohrer:
👍

PR #812, opened 2018-04-27
Add missing references to other Robot APIs

stefaniapedrazzi:
Usually we put the links to the other APIs in each function section, but some of them were missing.

### DIFF
--- a/reference/robot.md
+++ b/reference/robot.md
@@ -605,6 +605,8 @@ It does not matter whether the controller is synchronized or not.
 
 **wb\_robot\_task\_new** - *start a new thread of execution*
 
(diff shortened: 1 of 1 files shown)


stefaniapedrazzi:
:+1:

reference/robot.md
```diff
@@ -694,6 +696,8 @@ The message is received using the `webots.window("<robot window name>").receive`
 
 **wb\_robot\_window\_custom\_function** - *communication with the native C/C++ robot window [deprecated]*
 
+{[C++](cpp-api.md#cpp_robot)}, {[Java](java-api.md#java_robot)}, {[Python](python-api.md#python_robot)}, {[Matlab](matlab-api.md#matlab_robot)}, {[ROS](ros-api.md)}
+
```
omichel: Why not using `ros-api.md#robot` like for other languages?

stefaniapedrazzi: Probably the tags in the ROS api page have been added later after this was set.
All the references to the ROS API in any node section doesn't use the node tag.
We should open an issue for this.

Issue #816, opened 2018-05-01
Hashtag links are broken on Firefox/Windows


omichel:
https://www.cyberbotics.com/doc/guide/robotis-op2#symmetry-wbt doesn't scroll down to the tag and the index links are broken.

I believe this was broken recently it was working when @DavidMansolino posted this: https://www.cyberbotics.com/forum?message=8064

Maybe it was broken by #800?

stefaniapedrazzi:
I also noticed this morning on Linux/Firefox with:
https://www.cyberbotics.com/doc/guide/using-python#use-an-alternative-python-version

But on my Linux/Chrome it works correctly.

fabienrohrer:
I have the same on macOS, this is working on Chrome, but not on Firefox.

omichel:
By saving a doc page (see attachment) and analyzing it with https://validator.w3.org I found several errors that may explain the problem, in particular "Error: Attribute name not allowed on element h2 at this point.".

[doc.zip](https://github.com/omichel/webots-doc/files/1966193/doc.zip)

omichel:
Apparently hashtag links should refer to the `id` attribute in HTML5, not to the `name` attribute, which is obsolete in HTML5. See last answer of https://stackoverflow.com/questions/25813113/what-is-the-usage-of-the-name-attribute-in-h1-h6-tags

fabienrohrer:
The reason of the original issue is that wwi.css (which had been introduced with the robot component) is modifying `body.scroll-behavior`: https://github.com/omichel/webots/blob/master/resources/web/wwi/wwi.css#L2
Modifying body or any tag without using class or ids is a very bad practice because it render any CSS unmodular.

PR #817, opened 2018-05-02
Hotfix: broken anchors in firefox

fabienrohrer:
Fix https://github.com/omichel/webots-doc/issues/816

Preview: https://www.cyberbotics.com/doc/guide/robotis-op2?version=hotfix-anchors-for-firefox#symmetry-wbt

### DIFF
--- a/css/webots-doc.css
+++ b/css/webots-doc.css
@@ -37,6 +37,9 @@
   src: local("fonts/Roboto-Mono"), url("fonts/Roboto-Mono.ttf") format("truetype");
 }
 
+body {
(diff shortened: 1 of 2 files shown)


fabienrohrer:
I tried several solutions, and if I add the constraint to not modify wwi.css (which is wiser on master), I can only override this CSS rule.

I propose to accept this as-is, and to fix this in a cleaner way in develop (I will take care of this as soon as this PR is merged in develop) by removing this kind of general rules from wwi.css.

omichel:
:eyes:

fabienrohrer:
I'm still looking at a bug in the scroll computation.

fabienrohrer:
I will do the other patch I was speaking about in another PR.

fabienrohrer:
👍

PR #831, opened 2018-05-15
Improve API headers

### LINKED ISSUE #591: Language APIs docs are not user friendly

fabienrohrer:
Partially fix #591 (in prevision to do better later)

Aim: see the API functions in the page index.

Before: https://www.cyberbotics.com/doc/reference/camera
After: https://www.cyberbotics.com/doc/reference/camera?version=enhancement-api-headers

### DIFF
--- a/automobile/car-library.md
+++ b/automobile/car-library.md
@@ -4,9 +4,10 @@ The [car](#car-library) library is supposed to be used together with the [driver
 It provides additional information and functions to which a normal human driver of a car does not have access (e.g., changing the blinking period of the indicator or getting the value of the wheels encoders).
 All the functions included in this library are explained below.
 
-**Name**
+### Car Library Functions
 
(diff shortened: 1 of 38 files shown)


fabienrohrer:
Can I scale with this solution?

omichel:
Yes, but it doesn't look super nice when the function names are on two lines (the indentation of the second line indentation looks bad).

omichel:
Yes, it looks much better to me. Thank you.

fabienrohrer:
Ok, I will scale with this.

fabienrohrer:
I still improved the API. Mainly, the OO links are now displayed as tags:

capture d ecran 2018-05-16 a 12 10 03

fabienrohrer:
👍

fabienrohrer:
👍

PR #834, opened 2018-05-17
Language and OS toggles

### LINKED ISSUE #591: Language APIs docs are not user friendly

fabienrohrer:
Aim: Fix #591 by creating more dynamic pages where the target language or the OS (or more) may be selected through "radio buttons".


fabienrohrer:
As discussed this morning, a tab-based component is certainly simpler to use for this :-{(

Issue #180, opened 2016-09-08
Units and range in documentation


ghost:
_From @fabienrohrer on January 19, 2016 8:36_

Units and range in documentation are heterogenous, and sometimes missing.

Related forum messages:
https://www.cyberbotics.com/forum?message=5836
https://www.cyberbotics.com/forum?message=1339
https://www.cyberbotics.com/forum?message=2361

_Copied from original issue: omichel/webots#3414_

daniel-cyberbotics:
_From @fabienrohrer on January 19, 2016 8:36_

About fields, the reference manual follows more or less this rule:

```
NodeName {
  VRMLType fieldName defaultValues # [\[\(]minRange,maxRange[\]\)], (units)
}
```

All of the case can be seen here:
https://www.cyberbotics.com/reference/contactproperties.php

:arrow_right: I would propose to homogenize this.

Some of them are showing a small label explaining the field:
https://www.cyberbotics.com/reference/accelerometer.php
But this seems redundant with the description bellow.

:arrow_right: I would propose to drop this description label.

This is redundant also with the comments here:
https://github.com/omichel/webots/blob/master/resources/nodes/

:arrow_right: I would propose to remove all the comments here.

About function units, it is generally described in the comments.
For example, the return of the [m/s^2] of the [Accelerometer's wb_accelerometer_get_values() function](https://www.cyberbotics.com/reference/accelerometer.php#wb_accelerometer_get_values).

:arrow_right: I would propose to describe the units of the function into the function description.

daniel-cyberbotics:
_From @fabienrohrer on January 19, 2016 8:37_

If everyone agree, I can begin this refactoring.

daniel-cyberbotics:
_From @omichel on January 19, 2016 8:45_

I wouldn't drop the comments from the *.wrl files as they are useful when you look at these files to understand the nodes without having to go to the reference manual.

daniel-cyberbotics:
_From @omichel on January 19, 2016 8:49_

Maybe you address a single node for now and post the results here for review, so that we can agree if this is the right way to do.

daniel-cyberbotics:
_From @fabienrohrer on January 19, 2016 8:52_

> I wouldn't drop the comments from the *.wrl files as they are useful when you look at these files to understand the nodes without having to go to the reference manual.

Maintaining duplicate data should be avoided in every case. The proof of this is that this directory is currently completely unsynchronized from the reference manual.
The users are not interesting in seeing this directory.
[...]

PR #837, opened 2018-05-22
Fixed missing units for motor API

omichel:
Partial fix for #180 (regarding motors).

Test page: https://cyberbotics.com/doc/reference/motor?version=fix-motor-units

### DIFF
--- a/reference/motor.md
+++ b/reference/motor.md
@@ -17,7 +17,8 @@ Motor {
 
 ### Description
 
-A [Motor](#motor) node is an abstract node (not instantiated) whose derived classes can be used in a mechanical simulation to power a joint hence producing a motion along, or around, one of its axes.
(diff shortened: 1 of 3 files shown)


omichel:
👍

reference/motor.md
```diff
@@ -311,18 +315,22 @@ wb_motor_set_velocity(tag, desired_speed);  // rad/s
 The `wb_motor_get_target_position` function allows the user to get the target position.
 This value matches with the argument given to the last `wb_motor_set_position` function call.
 
-The `wb_motor_set_velocity` function specifies the velocity that motor should reach while moving to the target position.
+The `wb_motor_set_velocity` function specifies the velocity that a motor should reach while moving to the target position.
+The velocity is expressed in *radian per second* for rotational motors and in *meter per second* for linear motors.
```
stefaniapedrazzi: to be consistent we should maybe add the abbreviation also in case of the position, i.e. `(rad/s)` and `(m/s)`?
personally I found that the symbol is more immediate to understand than the extended text

same at line 30, 64, 270, 327

omichel:
👍

PR #841, opened 2018-05-23
Tab component

### LINKED ISSUE #591: Language APIs docs are not user friendly

fabienrohrer:
Fix #591 (second part)

### Aims:

- create the tab component
- migrate the C++, Java, Python, Matlab and ROS APIs to the "Nodes and API Functions" chapter.

### Preview:

https://www.cyberbotics.com/doc/reference/nodes-and-api-functions?version=feature-tab-component


fabienrohrer:
Could I have some feedback on the current state of this PR?

The idea is that a tab component can be created with this syntax:

    %tab-component

    %tab "Title 1"

    |                        |
    |------------------------|
    | It could be a table... |

    %tab-end

    %tab "Title 2"

    ```java
    // It could be code...
    ```

    %tab-end

    %end

Example:
- https://github.com/omichel/webots-doc/blob/ebe193fd0c41b3e960f643e02838bf032394ea78/reference/camera.md
- https://www.cyberbotics.com/doc/reference/camera?version=feature-tab-component#wb_camera_disable

Considerations:

- All the API described in the API pages can be moved in node pages.
- The tab component can be used in other contexts than languages.
- the tab content is set in full screen. This is working only for table and code for now, but it could be easily extended to other tags like images, notes, etc.

fabienrohrer:
Other consideration:

- Python syntax highlighter cannot work well in this case, because functions are not terminated by a trailing `:` and miss their declarations. This is difficult to solve.
- the OO classes looks like if they were incomplete. Adding a `// ...` before closing their definition seem important:

        class Camera : public Device {
          virtual void enable(int sampling_period);
          virtual void disable();
          int getSamplingPeriod();
          // ...
        }

fabienrohrer:
> > the tab content is set in full screen
> 
> I don't understand what you mean here...

I mean that the table or the code is not surrounded by margin or padding, overriding the general rules of the documentation.

fabienrohrer:
Ok, I fixed your comments.

I will start the scale tomorrow if there is nothing else.

omichel:
I am still a little bit unsure about the style of the tab buttons (especially the blue top ribbon), but I guess this can easily be changed later on.

fabienrohrer:
Yes, I can propose a more discreet and homogenous style later.
The main object to review is the mechanism, because the migration of the API is a huge task that I don't want to do twice or for nothing. The main thing to agree now is the modifications on the MD files.

omichel:
If I switch to Java for one component, shall we switch to Java all other components? I think yes (not sure)... Also, once I switched to Java one component and I visit another page, I believe all the components on this new page should be on the Java tab. Isn't it?

omichel:
I prefer the new style to the previous one (it's lighter and better matches the color style of the site). Thanks!

DavidMansolino:
It is very nice, I really like this new way of documenting the other API. I just have a minor comment, for ROS (for example in https://www.cyberbotics.com/doc/reference/camera?version=feature-tab-component#camera-functions), it is not very clear that some 'data type' are links, furthermore, I am not sure it is good to put 'N/A' because it is not the case, it is just that we don't repeat the definition for common services.

DavidMansolino:
Before scaling, by careful, it seems you always have a typo in C++ header, there is an unwanted  '`'.

omichel:
Also, in the data type column, can we remove the `webots_ros::` namespace instead of repeating it for every data type? In such a case we should have also change the header of this column from "data type" to "ROS data type".

DavidMansolino:
>Also, in the data type column, can we remove the webots_ros:: namespace instead of repeating it for every data type?

Note that not all of them are using the `webots_ros` namespace, but some of them (those using default message definition) use for for example the `sensor_msgs` namespace.

omichel:
> Note that not all of them are using the webots_ros namespace, but some of them (those using default message definition) use for for example the sensor_msgs namespace.

OK, I was not sure about it. So, in that case forget my comment about it.

fabienrohrer:
Thank you for all your pertinent comments. I already fixed obvious issues.

> If I switch to Java for one component, shall we switch to Java all other components? I think yes (not sure)... Also, once I switched to Java one component and I visit another page, I believe all the components on this new page should be on the Java tab. Isn't it?

It seems indeed good ideas. I would propose to implement this later (after merging this PR?) in order to facilitate the comparison between the 2 possibilities. 

> Also for the ROS table, maybe the "topic/service" column could be narrowed to be able to enlarge the "name" column.

This would be indeed better but it's unfortunately tricky to fix in a clean way. I will give a try.

fabienrohrer:
👍

omichel:
:eyes:

omichel:
The text of the ROS table is overflowing the table:
image

reference/lidar.md
```diff
@@ -54,14 +58,60 @@ typedef struct {
...
+  public float getZ();
+  public float getTime();
+  public double getLayerId();
+}
+```
+
+%tab-end
+
+%end
+
```
omichel: I believe we miss a MATLAB and a ROS tab here. If a structure/class is not available, we should at least provide some plain text explanation that document how to achieve the same in these APIs.

reference/lidar.md
```diff
@@ -149,6 +201,75 @@ void wb_lidar_disable(WbDeviceTag tag);
...
+```
+
+%tab-end
+
+%tab "ROS"
+
+| name | service/topic | data type | data type definition |
+| --- | --- | --- | --- |
+| `/<device_name>/enable` | `service` | [`webots_ros::set_int`](ros-api.md#common-services) | |
+| `/<device_name>/get_sampling_period` | `service` | [`webots_ros::get_int`](ros-api.md#common-services) | |
```
omichel: Generally, in the ROS table, the hyperlinks are not visible unless you move the mouse over them and see that you can click them. I believe we should make them more visible (using a CSS rule) otherwise, users will miss them.

fabienrohrer:
@omichel I think I addressed all your reported issues.

👍

omichel:
👍

stefaniapedrazzi:
:eyes:

reference/supervisor.md
```diff
@@ -122,28 +163,335 @@ If no node is currently selected, the function returns NULL.
...
+
+%tab "C++"
+
+```cpp
+#include <webots/Node.hpp>
+
+namespace webots {
+  class Node {
+    enum {
+      NO_NODE, ACCELEROMETER, APPEARANCE, BACKGROUND, BALL_JOINT, BALL_JOINT_PARAMETERS,
```
stefaniapedrazzi: The list of nodes is not up-to-date (same for Java and Python) but we can update it on a different PR.

missing:
* WB_NODE_LENS
* WB_NODE_LENS_FLARE
* WB_NODE_MUSCLE
* WB_NODE_MULTI_TEXTURE
* WB_NODE_COMPOSED_CUBE_MAP_TEXTURE
* WB_NODE_COMPOSED_SHADER
* WB_NODE_SKIN
* WB_NODE_UNIFORM

to be removed:
* WB_LENS_DISTORTION

reference/supervisor.md
```diff
@@ -28,58 +28,106 @@ For functions returning a string, an empty string is returned instead of a NULL
...
+
+%tab-end
+
+%tab "MATLAB"
+
+```matlab
+node = wb_supervisor_node_get_root()
+node = wb_supervisor_node_get_self()
+node = wb_supervisor_node_get_from_def('def')
+node = wb_supervisor_node_get_from_id('id')
```
stefaniapedrazzi: `id` argument is not a string, so it should probably not be quoted

reference/speaker.md
```diff
@@ -94,6 +221,86 @@ const char *wb_speaker_get_language(WbDeviceTag tag);
...
+
+%tab "ROS"
+
+| name | service/topic | data type | data type definition |
+| --- | --- | --- | --- |
+| `/<device_name>/get_engine` | `service` | [`webots_ros::get_string`](ros-api.md#common-services) | |
+| `/<device_name>/get_language` | `service` | [`webots_ros::get_string`](ros-api.md#common-services) | |
+| `/<device_name>/set_engine` | `service` | [`webots_ros::set_string`](ros-api.md#common-services) | |
+| `/<device_name>/set_language` | `service` | [`webots_ros::set_string`](ros-api.md#common-services) | |
+| `/<device_name>/speak` | `service` | webots_ros::speaker\speak | `string text`<br/>`float64 volume`<br/>`---`<br/>`int8 success`<br/> |
```
stefaniapedrazzi: "webots_ros::speaker\speak" -> "webots_ros::speaker\\_speak"

reference/speaker.md
```diff
@@ -16,14 +16,77 @@ It can be used to play sounds and perform text-to-speech from the controller API
...
+%tab-end
+
+%tab "Python"
+
+```python
+from controller import Speaker
+
+class Speaker (Device):
+    @staticmethod
+    def playSound(left, right, sound, volume, pitch balance, loop):
```
stefaniapedrazzi: missing comma between pitch and balance

reference/robot.md
```diff
@@ -215,110 +284,195 @@ int main() {
...
+tag = wb_robot_get_device('name')
 ```
 
+%tab-end
+
+%tab "ROS"
+
+| name | service/topic | data type | data type definition |
+| --- | --- | --- | --- |
+| `/robot/time_step` | `service` | [`webots_ros::set_int`](ros-api.md#common-services) | |
```
stefaniapedrazzi: This function has been copied in the wrong section.
The ROS API doesn't have any method to get the device because it doesn't need it.

reference/robot.md
```diff
@@ -126,6 +129,72 @@ void wb_robot_init();
...
+%tab "C++"
+
+```cpp
+#include <webots/Robot.hpp>
+
+namespace webots {
+  class Robot : public Device {
+    Robot();
+    virtual ~Robot();
+    virtual int step(int sampling_period);
```
stefaniapedrazzi: it would be better to use `duration` (as in C API) instead of `sampling period` also for the C++, Java, Python and MATLAB

reference/robot.md
```diff
@@ -378,6 +599,78 @@ double wb_robot_battery_sensor_get_value();
...
+value = wb_robot_battery_sensor_get_value()
+```
+
+%tab-end
+
+%tab "ROS"
+
+| name | service/topic | data type | data type definition |
+| --- | --- | --- | --- |
+| `/battery_sensor/value` | `topic` | webots_ros::Float64Stamped | [`Header`](http://docs.ros.org/api/std_msgs/html/msg/Header.html) `header`<br/>`float64 data` |
```
stefaniapedrazzi: Here we miss the `/battery_sensor/enable` and  `/battery_sensor/get_sampling_period` services documentation

reference/touchsensor.md
```diff
@@ -150,27 +232,97 @@ This function can be used with a sensor of type "force-3d" exclusively.
...
+}
+```
+
+%tab-end
+
+%tab "MATLAB"
+
+```matlab
+WB_TOUCH_SENSOR_BUMPER, WB_TOUCH_SENSOR_FORCE, WB_TOUCH_SENSOR_FORCE3D
+type = wb_touch_sensor_get_type(tag)
```
stefaniapedrazzi: in other device pages there is a space after the constant values

fabienrohrer:
Note: #848 should not be forgot when applying this PR to develop.

reference/supervisor.md
```diff
@@ -122,28 +163,335 @@ If no node is currently selected, the function returns NULL.
...
+  // ...
+}
+```
+
+%tab-end
+
+%tab "MATLAB"
+
+```matlab
+type = wb_supervisor_node_get_type(node)
```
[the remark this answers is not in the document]
fabienrohrer: No, this is an issue indeed.

reference/supervisor.md
fabienrohrer: I would prefer in another PR, it would be time to recheck all this topic, and possibly add tests for this.

fabienrohrer:
I have to applied the tab-component on the automobile library pages too..

fabienrohrer:
Since last review:
- I fixed all the Stefania and Olivier comments.
- I applied tab components on [Driver](https://www.cyberbotics.com/doc/automobile/driver-library?version=feature-tab-component) and [Car](https://www.cyberbotics.com/doc/automobile/car-library?version=feature-tab-component) libraries
- I removed the OO tags feature which was no more used and which was causing #850 : https://www.cyberbotics.com/doc/guide/programming-fundamentals?version=feature-tab-component
- I proofread the modified pages once.

👍

stefaniapedrazzi:
Some of my comments in supervisor.md have not been fixed and there are still errors in other pages that I didn't review.
But for me it is fine and I will check again all the nodes pages and fix the errors I see once this PR is fixed.

I will just check the modifications on the automobile documentation.

:eyes:

automobile/car-library.md
```diff
@@ -28,13 +89,88 @@ If you use the [driver](driver-library.md) library it is not needed to call thes
...
 
+typedef enum {
+  WBU_CAR_TRACTION,
+  WBU_CAR_PROPULSION,
+  WBU_CAR_FOUR_BY_FOUR
+} wbu_car_type;
+
 wbu_car_type wbu_car_get_type();
-wbu_car_engine_type wbu_car_get_type();
+wbu_car_engine_type wbu_car_get_engine_type();
```
stefaniapedrazzi: `wbu_car_engine_type` enum is not declared here for the C API

automobile/car-library.md
```diff
@@ -130,13 +454,81 @@ All these functions provide important physical characteristics from the car.
...
+%tab-end
+
+%tab "C++"
+
+```cpp
+#include <webots/Car.hpp>
+
+namespace webots {
+  class Car : public Driver {
+    double getWheelEncoder(int wheel_index);
```
stefaniapedrazzi: missin declaration of enum for WHEEL_FRONT_RIGHT, etc.

same for Python and Java.

fabienrohrer:
👍

PR #843, opened 2018-05-23
Robot component: Fix device name wrap

### LINKED ISSUE #842: Robot compenent: right menu size doesn't adapt to long device names
It seems that the robot component doesn't adapt to long device names: the device names are written outside the grey box.

on Linux/Chrome:
robot_component

On Firefox/Linux the appearance is the same, i.e. the grey background is smaller than the device name, but there is an horizontal scrollbar so that the whole device name can be read.

fabienrohrer:
Fix #842
Preview: https://www.cyberbotics.com/doc/guide/pr2?version=fix-robot-component-name-wrap

### DIFF
--- a/css/webots-doc.css
+++ b/css/webots-doc.css
@@ -548,6 +548,7 @@ body {
 .robot-component .device-name {
   color: white;
   font-family: roboto;
+  word-wrap: break-word;
 }
 .robot-component .device {
   background-color: #666;


fabienrohrer:
👍

stefaniapedrazzi:
In this new version, the device name is cut on multiple lines:
for example just the last "r" of `bl_caster_rotation_joint_sensor` is written on a new line.
Is this the expected behavior?

fabienrohrer:
Yes, this is annoying but expected. Sorry for the lack of explanations.

I have no much better solution to propose:

- either the menu width could be increased. in this case, the issue is just reported.
- either the menu width could be adaptative. in this case, menu width is not fixed and could become huge and hide the 3D model.
- either the font could be reduced. it would looks very weird in the page, and the issue is just reported.
- either an horizontal scroll bar can be added, but it will break the modern rendering of the component.
- either we break the device name into words (as proposed here), with the issues you mentioned.. Note there is no better CSS text-wrap rule.

PR #860, opened 2018-06-06
Remove wrong superclass from Joystick and Keyboard classes

fabienrohrer:
Joystick and Keyboard does not inherits from Device:

https://github.com/omichel/webots/blob/master/include/controller/cpp/webots/Keyboard.hpp
https://github.com/omichel/webots/blob/master/include/controller/cpp/webots/Joystick.hpp

This is already correct for the Mouse class documentation:
https://www.cyberbotics.com/doc/reference/mouse#wb_mouse_enable

### DIFF
--- a/reference/joystick.md
+++ b/reference/joystick.md
@@ -36,7 +36,7 @@ int wb_joystick_get_sampling_period();
 #include <webots/Joystick.hpp>
 
 namespace webots {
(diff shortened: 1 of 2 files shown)


fabienrohrer:
👍

DavidMansolino:
👀

PR #862, opened 2018-06-06
Fix VRML97 in Node chart

DavidMansolino:
https://www.cyberbotics.com/doc/reference/node-chart?version=fix-vrml97


DavidMansolino:
👍

PR #861, opened 2018-06-06
Let the doc match with the real nodes.h (and derived files) state

fabienrohrer:
This hotfix aim is to have a consistent state for the documentation, even if the list of the nodes in the API is not correct, and to fix this better for the next release after.

Preview: https://www.cyberbotics.com/doc/reference/supervisor?version=hotfix-node-list#wb_supervisor_node_get_base_type_name

PR #865, opened 2018-06-07
Cleanup API Enum/Define

DavidMansolino:
Dual PR: omichel/webots#7324


DavidMansolino:
Remains the list of node type to update, but I will wait for #861 to be merged.

reference/supervisor.md
```diff
@@ -399,9 +401,9 @@ namespace webots {
       HINGE_2_JOINT_PARAMETERS, IMMERSION_PROPERTIES, JOINT_PARAMETERS, LENS,
       LENS_FLARE, PHYSICS, RECOGNITION, SLIDER_JOINT, SOLID, SOLID_REFERENCE,
       SLOT, TRACK, TRACK_WHEEL, WORLD_INFO, ZOOM
-    };
+    } Type;
```
DavidMansolino: I think this new way of documenting the node list introduced in #861 is very missleading for the user as it doesn't reflect the actual order in the header, and order in an enumeration is **really** important...

stefaniapedrazzi: Why? Users don't need to know the value of the constant and don't need to use it. They just need to know the constant name.

stefaniapedrazzi: Ok, I understand what you mean.
I think that it is also very important to be able to find easily a node in a list, otherwise it is useless.
But probably we should not document it as an enumeration.
Or we should change the order in nodes.h, because the current one is just a big mess.

DavidMansolino: Because for example a user may want to use such kind of code to detect if the node is a device (there are of course other ways to know if it is a device, but this is just an example):
```
if (node->getType() >= Node::ACCELEROMETER && node->getType() <= Node::TOUCH_SENSOR)
  std::cout << This is a device ... or not maybe not. << std::endl;
```

DavidMansolino: > Or we should change the order in nodes.h, because the current one is just a big mess.

Yes I think that would be the best solution, because currently the order in nodes.h is completely right (it even doesn't respect our grouping rule).

DavidMansolino: Reported in omichel/webots#7420

DavidMansolino:
👍

automobile/driver-library.md
```diff
@@ -986,10 +986,10 @@ wbu_wipers_mode wbu_driver_get_wipers_mode();
 
 namespace webots {
   class Driver {
-    enum {DOWN, SLOW, NORMAL, FAST};
+    typedef enum {DOWN, SLOW, NORMAL, FAST} WhipersMode;
 
-    void setWipersMode(int mode);
-    int getWipersMode();
+    void setWipersMode(WhipersMode mode);
+    WhipersMode getWipersMode();
```
omichel: Typo: "Wiper" instead of "Whipers"

DavidMansolino:
👍

omichel:
Still some whiper with a "H"...

DavidMansolino:
👍

PR #871, opened 2018-06-11
Fix Vehicle typedef Names.

DavidMansolino:
Dual PR omichel/webots#7432


DavidMansolino:
Not sure whether we want to call them WbCar.../WbDriver... or WbuCar/WbuDriver... I have a slight preference for the second option which seems more coherent with the rest of the vehicle API.
@All, what yo you think?

stefaniapedrazzi:
Given that we are changing them, I would personally address the issue globally and make a "final" decision on how we want to handle all the utility libraries, i.e. address https://github.com/omichel/webots/issues/4134 not only for the types but also for functions.

DavidMansolino:
I agree it is good to take a decision about this and stick to it but I would like to keep this PR focused on typedef names only (especially because the dual PR is focused on this and because according to omichel/webots#4134 libcar and libdriver are already correct).

stefaniapedrazzi:
So then the type cannot be `WbCar.../WbDriver..` because only libController functions and types should start with `Wb`.
If we decide that all the utility libraries have to be external and use prefix `wbu`, then all the libCar/libDriver types should have the `Wbu` prefix and not `Wb`.

omichel:
Yes, the `Wbu` prefix is the way to go to solve this issue.

DavidMansolino:
👍

PR #872, opened 2018-06-11
Document New Mouse Functions

DavidMansolino:
dual PR: omichel/webots#7409

https://www.cyberbotics.com/doc/reference/mouse?version=feature-new-mouse-functions

### DIFF
--- a/reference/mouse.md
+++ b/reference/mouse.md
@@ -123,7 +123,7 @@ The `u` coordinate goes from 0 on the left border of the 3D window to 1 on the r
 These values may be `NaN` if the mouse is outside of the 3D window.
 
 The `x`, `y` and `z` fields are indicating the 3D coordinate where the mouse is pointing in the 3D window.
-These values may be `NaN` if not applicable, for example when the mouse is pointing to the scene background.
+These values may be `NaN` if not applicable, for example when the mouse is pointing to the scene background or when the 3d position is not enabled (see the [`wb_mouse_enable_3d_position` function](#wb_mouse_enable_3d_position)).
 
 ### Mouse Functions
 
--- a/reference/mouse.md
+++ b/reference/mouse.md
@@ -317,4 +317,4 @@ enabled = wb_mouse_is_3d_position_enabled()
 
 In order to retrieve the 3D coordinate where the mouse is pointing in the 3D window, a picking is necessary, this picking depending on the scene and the time step can considerably reduce the simulation speed.
 Therefore, to avoid  unnecessary computation due to the picking, it should be explicitly enabled using the `wb_mouse_enable_3d_position` function, otherwise the `x`, `y` and `z` fields of [`WbMouseState`](#wbmousestate) are set to `NaN`.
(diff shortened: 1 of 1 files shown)


DavidMansolino:
👍

omichel:
:eyes:

reference/mouse.md
```diff
@@ -226,3 +226,95 @@ The `sampling_period` parameter is expressed in milliseconds, and defines how fr
...
+
+```cpp
+#include <webots/Mouse.hpp>
+
+namespace webots {
+  class Mouse {
+    void enable3dPosition();
+    void disable3dPosition();
+    bool is3dPositionEnabled() const;
+  }
```
omichel: When not all methods are included, we use to to add a line like:
```c++
// ...
```
This is use many times in the reference manual, we should keep consistent with this (see for example the Accelerometer API).

reference/mouse.md
```diff
@@ -226,3 +226,95 @@ The `sampling_period` parameter is expressed in milliseconds, and defines how fr
...
+%tab "C++"
+
+```cpp
+#include <webots/Mouse.hpp>
+
+namespace webots {
+  class Mouse {
+    void enable3dPosition();
+    void disable3dPosition();
+    bool is3dPositionEnabled() const;
```
fabienrohrer: In C++, Java and Python, I added `// ...` or `# ...` in other APIs to suggest that the list of class functions is not exhaustive. `wb_mouse_enable` functions should be also updated according to this rule because of this addition.

reference/mouse.md
```diff
@@ -315,6 +315,8 @@ enabled = wb_mouse_is_3d_position_enabled()
 
 *enable the 3d position*
 
-In order to retrieve the 3D coordinate where the mouse is pointing in the 3D window, a picking is necessary, this picking depending on the scene and the time step can considerably reduce the simulation speed.
-Therefore, to avoid  unnecessary computation due to the picking, it should be explicitly enabled using the `wb_mouse_enable_3d_position` function, otherwise the `x`, `y` and `z` fields of [`WbMouseState`](#wbmousestate) are set to `NaN`.
-Once enabled, it is possible to disable it using the `wb_mouse_disable_3d_position` function, furthermore, it is at any time possible to check if it is enabled using the `wb_mouse_is_3d_position_enabled` function.
+In order to retrieve the 3D coordinate where the mouse is pointing in the 3D window, a picking is necessary.
+Depending on the complexity of the scene and time step, this picking can considerably reduce the simulation speed.
+Therefore, to avoid  unnecessary computation due, picking should be explicitly enabled using the `wb_mouse_enable_3d_position` function, otherwise the `x`, `y` and `z` fields of [`WbMouseState`](#wbmousestate) are set to `NaN`.
```
DavidMansolino: It seems there is an issue in your correction here: unnecessary computation **due**, picking
I think the 'due' should be removed, right?

omichel: Well spotted. I just fixed it.

DavidMansolino:
👍

Issue #870, opened 2018-06-08
Display with Attached Camera Should Disable Filtering


DavidMansolino:
We should document the fact that it is highly recommended to disable the ImageTexture filtering for display attached to a camera.

See https://github.com/omichel/webots/pull/7421#discussion_r194083671

PR #877, opened 2018-06-13
Document a recommendation about ImageTexture.filtering for Display

fabienrohrer:
Fix: https://github.com/omichel/webots-doc/issues/870

Preview: https://www.cyberbotics.com/doc/reference/display?version=enhancement-camera-filtering-documentation#description


fabienrohrer:
👍

fabienrohrer:
👍

tn12787:
It is not only for sharp display images - mipmapping needs to be explicitly disabled, or attached camera images will fade out as you zoom!

fabienrohrer:
Yes I know, but my concern was to find a very simple term to explain this to the user. I think if a user read this, it's because it has a blurry texture. I think only few users knows what is mipmap or anysotropic filtering.
Maybe I can add this in parenthesis.

fabienrohrer:
👍 

Is it better like this?

fabienrohrer:
👍

Issue #879, opened 2018-06-13
Wrong paths to Automobile libraries in the Using Visual C++ with Webots page


DavidMansolino:
The paths here are wrong: https://www.cyberbotics.com/doc/guide/using-visual-cpp-with-webots#link-with-the-webots-libraries

DavidMansolino:
Fixed in #883

PR #883, opened 2018-06-14
Fix Vehicle Libraries Path

DavidMansolino:
Fix issue #879
Preview: https://www.cyberbotics.com/doc/guide/using-visual-cpp-with-webots?version=fix-vehicle-libraries-path-to-master#link-with-the-webots-libraries


DavidMansolino:
👍

PR #887, opened 2018-06-14
Fix alphabetical order of nodes in list

stefaniapedrazzi:
I just found an additional unordered node.
Webots repository needs to be fixed accordingly.


stefaniapedrazzi:
:+1:

PR #888, opened 2018-06-15
Add a node about physical behaviors of non top robot nodes

stefaniapedrazzi:
Try to document why a Robot inserted in another Robot node doesn't behave the same way as if it would be a top node.

This is maybe not necessary because it follows the same rules as for any Solid node.
But given that robots could move I think it is not bad to explicitly write it again.


stefaniapedrazzi:
:+1:

reference/robot.md
```diff
@@ -21,9 +21,11 @@ Direct derived nodes: [DifferentialWheels](differentialwheels.md), [Supervisor](
 
 ### Description
 
-The [Robot](#robot) node can be used as basis for building a robot, e.g., an articulated robot, a humanoid robot, a wheeled robot...
+The [Robot](#robot) node can be used as basis for building a robot, e.g., an articulated robot, a humanoid robot, a wheeled robot.
 If you would like to build a robot with supervisor capabilities use the [Supervisor](supervisor.md) node instead (Webots PRO license required).
 
+> **Note**: If the Robot node has one or more Solid (or derived) ancestor nodes, then the physical properties of the ancestor nodes will affect the Robot node physical behavior.
```
omichel: I would this sentence with "Obviously": "Obviously, if the Robot node..."

tn12787: "Obviously" is very harsh like this. I would say "Logically".

stefaniapedrazzi:
:+1:

PR #896, opened 2018-06-20
Update tutorials

DavidMansolino:
Update the tutorials according to the user interface changes in R2018b.


DavidMansolino:
:+1:    (the rest of the tutorials seems still valid).

Issue #897, opened 2018-06-20
Replace 'Main Toolbar' decsription by 'Context Menu' description


DavidMansolino:
This section should be removed/shortened and instead we should add a section about the context menu: https://www.cyberbotics.com/doc/guide/the-user-interface#main-toolbar

DavidMansolino:
Fixed in #899

PR #899, opened 2018-06-21
Update User Interface

DavidMansolino:
Fix issue #897 

Previews:
  - https://www.cyberbotics.com/doc/guide/the-3d-window?version=update-user-interface#selecting-an-object
  - https://www.cyberbotics.com/doc/guide/the-user-interface?version=update-user-interface#main-toolbar


guide/the-3d-window.md
```diff
@@ -5,8 +5,7 @@
 A single mouse click allows you to select a solid object.
 The bounding object of a selected solid is represented by white lines.
 These lines turn pink if the solid is colliding with another one and blue when the solid is idle, i.e., it comes to rest and it does not interact with any other active solid.
-Selecting a robot enables the `Show Robot Window` item in the `Tools` menu.
-Double-clicking on a solid object opens the Robot Window and if it was closed, the Scene Tree.
```
DavidMansolino: > Double-clicking on a solid object opens [...] if it was closed, the Scene Tree.

Was this even working at some point? It doesn't seems to be the case neither on master or develop.

stefaniapedrazzi: Yes, I think that at some point (early Webots 7 or previous vesions) this was working.
In any case this is no longer the case now.

DavidMansolino:
👍

guide/the-3d-window.md
DavidMansolino: Ok thank you, I will remove it for now and open an issue on Webots repository to discuss if we want to restore this feature or not.

stefaniapedrazzi:
:eyes:

guide/the-3d-window.md
```diff
@@ -2,15 +2,25 @@
...
 
+#### Context Menu
+
+%figure "The context menu allows you to perform various actions depending on the current selected object"
+
+![context_menu.png](images/context_menu.png)
+
+%end
+
+When an object is selected in the 3D Window is is possible to right click on it to open the context menu.
```
stefaniapedrazzi: An object doesn't need to be selected to open the context menu.
It is the right-click action that picks the object and select it before opening its context menu.

In the rest of the page we use `3D window`

typo: "is is possible" -> "it is possible"

guide/the-3d-window.md
```diff
@@ -2,15 +2,25 @@
...
+This menu allows you to perform various actions such as: copy the object, delete the object, move the viewpoint to the object and show the robot window (if the object is a [Robot](../reference/robot.md)).
```
stefaniapedrazzi: I'm not sure that it is correct to use the colon after "such as".
Here (Other tips section) for example they explicitly suggest to not do it:
https://www.aje.com/en/arc/editing-tip-such-and-including/

guide/the-user-interface.md
```diff
@@ -369,8 +369,7 @@ The remaining menu items bring up various information as indicated, in the form
 
 ### Main Toolbar
 
-The main toolbar contains items for editing the world and the speedometer (see this [subsection](#speedometer-and-virtual-time)), other than shortcuts to items of the `File`, `Simulation` and `View` menus.
-Edit actions always apply on the selected object, that is highlighted both in the 3D window and in the Scene Tree.
+The main toolbar contains items for adding new nodes to the world, the speedometer (see this [subsection](#speedometer-and-virtual-time)) and shortcuts to items of the `File`, `Simulation` and `View` menus.
```
stefaniapedrazzi: Note that "items for adding new nodes" is actually only one item.

DavidMansolino:
👍

PR #902, opened 2018-06-22
Cleanup requried package for physic plugins

DavidMansolino:
Since it is not possible anymore to call openGl directly from physics plugin I think these packages are not required anymore.


DavidMansolino:
:+1:

stefaniapedrazzi:
The physics plugin still links to the GL libraries:
https://github.com/omichel/webots/blob/8b340ace515fbb166591a634012ce69a83382c68/include/plugins/physics.h#L11
https://github.com/omichel/webots/blob/8b340ace515fbb166591a634012ce69a83382c68/resources/Makefile.include#L386


First these dependencies have to be removed, otherwise in the current status these libraries are still needed to compiled physics plugins and controller plugins.

DavidMansolino:
:+1:   (fixed conflict)

PR #920, opened 2018-07-04
Correct some obsolete blog info

tn12787:
Fix some info in the new blog post.

### DIFF
--- a/blog/Webots-2018-b-release.md
+++ b/blog/Webots-2018-b-release.md
@@ -67,7 +67,7 @@ We plan to continue developing advanced graphics features for Webots in the comi
 We've done lots of work on user experience with this latest update.
 Shown above is one of the new looks for Webots R2018b, Webots Night.
 This new look features clean lines, flat design, and a subdued color palette with highlights in just the right places to make interacting with Webots even more intuitive.
-If you prefer the old look, we've also kept the previous style, in a theme called "Webots Classic", enabled by default.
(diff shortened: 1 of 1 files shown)


tn12787:
👍

blog/Webots-2018-b-release.md
```diff
@@ -65,9 +65,9 @@ We plan to continue developing advanced graphics features for Webots in the comi
 %end
 
 We've done lots of work on user experience with this latest update.
-Shown above is the new default look for Webots R2018b, Webots Night.
+Shown above is one of the new looks for Webots R2018b, Webots Night.
 This new look features clean lines, flat design, and a subdued color palette with highlights in just the right places to make interacting with Webots even more intuitive.
-If you prefer the old look, we've also kept the previous style, in a theme called "Webots Classic".
+If you prefer the old look, we've also kept the previous style, in a theme called "Webots Classic", enabled by default.
```
DavidMansolino: Actually, there isn't a default theme, the user has to chose which theme he wants to use at first startup

tn12787: You're right, I figured that given that the dialog itself is in the classic theme it makes sense to say it's "enabled by default", but this probably isn't clear enough.

DavidMansolino: Yes you are right the dialog uses the classic theme by default but not the rest of Webots.

tn12787:
👍

PR #939, opened 2018-07-31
Fix function name consistency between code and documentation

stefaniapedrazzi:
Argument name in documentation doesn't match the one in the code:
``webots/include/plugins/physics.h``


stefaniapedrazzi:
:+1:

PR #940, opened 2018-08-21
Port recent changes to Webots-doc

DavidMansolino:
The goal of this PR is to port the changes recently done in omichel/webots-dev@master into this repo in order to:
  - Fix urgent issue with main readme file.
  - Re-initialize the doc subrepo from omichel/webots-dev in order to try to have a clean/working subrepo workflow.

PR #942, opened 2018-08-22
Fixed merge issue.

DavidMansolino:
Merge issue from  #940, because these modification were present in webots-doc but not in webots-dev.


DavidMansolino:
👍

PR #946, opened 2018-10-31
Feature doc tab component selection

DavidMansolino:
This PR adds the possibility to set which tab should be opened by default when opening a doc page.
This will be very helpful when referring to a specific function in a specific language to a user.

This PR targets this repo for testing purpose only.

### DIFF
--- a/doc.php
+++ b/doc.php
@@ -19,14 +19,16 @@
   $uri = substr($request_uri, 5); // we remove the "/doc/" prefix
   $i = strpos($uri, '/');
   unset($repository);
+  $branch = '';
+  $tab = '';
   if ($i !== FALSE) {
     $book = substr($uri, 0, $i);
     $j = strpos($uri, 'version=');
     if ($j === FALSE) {
       $page = substr($uri, $i + 1);
-      $branch = '';
       $n = strpos($uri, 'tab=');
-      $tab = substr($uri, $j + 4);
+      if ($n !== FALSE)
+        $tab = substr($uri, $j + 4);
     } else {
       $page = substr($uri, $i + 1, $j - $i);
       $version = substr($uri, $j + 8);
@@ -47,8 +49,6 @@
     # default values:
     $book = $uri;
     $page = 'index';
-    $branch = '';
-    $tab = ''
     # anchor is not sent to the server, so it has to be computed by the javascript
   }
   if (!isset($repository))
--- a/doc.php
+++ b/doc.php
@@ -28,13 +28,13 @@
       $page = substr($uri, $i + 1);
       $n = strpos($uri, 'tab=');
       if ($n !== FALSE)
-        $tab = substr($uri, $j + 4);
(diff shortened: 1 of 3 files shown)


DavidMansolino:
👍   This PR is ready for review, unfortunately it can be tested only locally because online the 'doc.php' file need to be updated.

doc.php
```diff
@@ -15,19 +15,26 @@ function endsWith($haystack, $needle) {
     header("Location: $request_uri");
     exit();
   }
-  # the URL follow this format https://www.cyberbotics.com/doc/book/page?version=tagOrBranch#anchor where version and anchor are optional
+  # the URL follow this format https://www.cyberbotics.com/doc/book/page?version=tagOrBranch&tab=C#anchor where version, tab and anchor are optional
```
omichel: Instead of `tab`, I would call this parameter "programming language", but because it's a bit long, a better name could be `api`. What do you think?

omichel: Also, we should list here all the valid values for this new parameter, for C or Python, it is obvious, but I guess that for C++, we cannot write `tab=C++` as the `+` character is a separator... So what shall we write for C++? `cpp` or `c%43%43`?

DavidMansolino: > Instead of tab, I would call this parameter "programming language", but because it's a bit long, a better name could be api. What do you think?

As just discussed in the office, the tab concept is much more global than just language, currently we use it only for language but nothing in the code prevents us from using for anything else.

> Also, we should list here all the valid values for this new parameter, for C or Python, it is obvious, but I guess that for C++, we cannot write tab=C++ as the + character is a separator... So what shall we write for C++? cpp or c%43%43?

I just tested locally in Chrome, Firefox and Edge and it is working fine, I will test again locally once doc.php is updated.

DavidMansolino:
👍

DavidMansolino:
This new feature can now be tested online: https://www.cyberbotics.com/doc/reference/accelerometer?version=feature-doc-tab-component-selection&tab=python#accelerometer-functions

There is just one issue I need to fix, it is that when changing tab the menu urls are not updated.

DavidMansolino:
👍

fabienrohrer:
Sorry to come late, this PR is really good but this url is currently not working:

https://www.cyberbotics.com/doc/reference/supervisor?tab=c++

=> the index page is displayed.

This error is displayed:

```
Erreur d’analyse XML : erreur de syntaxe
Emplacement : https://raw.githubusercontent.com/omichel/webots-doc/master/reference/supervisor?tab=c++.md
Numéro de ligne 1, Colonne 1 :
```

DavidMansolino:
Thank you for the feedback, I will fix this.
Unfortunately this was not reproducible before merging this PR because locally or with a version (e.g. https://www.cyberbotics.com/doc/reference/supervisor?version=master&tab=python#wb_supervisor_field_insert_mf_vec2f) it is working fine.

PR #947, opened 2018-11-06
Update canonical URL when navigating

omichel:
This should improve google indexing.
Test page: https://cyberbotics.com/doc/reference/index?version=fix-update-canonical-url-when-navigating

### DIFF
--- a/js/viewer.js
+++ b/js/viewer.js
@@ -572,6 +572,8 @@ function updateBrowserUrl() {
     } catch (err) {
     }
   }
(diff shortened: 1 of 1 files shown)


omichel:
Updating the canonical URL this way is tricky and I am not sure it will be seen by the google crawler, but we should give it a try. If that doesn't work, we should have a different generated content on the HTML page, e.g., use PHP instead of Javascript to generate the page content.

👍

PR #948, opened 2018-11-06
Update dependencies.txt

omichel:
- Uses CDNs instead of Cyberbotics URLs for a better google indexing.
- Removed duplicate `jquery` and `jqueryui` includes from the web.
- Added `jquery` and `jqueryui` dependencies in a local dependencies file.
- Removed useless `.map` files.
- Fixed wrong `jquery-ui` folder name to `jqueryui` (needs to be fixed in Webots too).


omichel:
👍 (note: Webots will be fixed in a separate PR for `jquery-ui` folder name change to `jqueryui`).

local_dependencies.txt
```diff
@@ -0,0 +1,23 @@
+# jquery
+
+https://cdnjs.cloudflare.com/ajax/libs/jquery/1.11.3/jquery.min.js
+
+# jqueryui
+
+https://cdnjs.cloudflare.com/ajax/libs/jqueryui/1.11.4/jquery-ui.min.css
+https://cdnjs.cloudflare.com/ajax/libs/jqueryui/1.11.4/jquery-ui.min.js
+https://cdnjs.cloudflare.com/ajax/libs/jqueryui/1.11.4/jquery-ui.min.css
+https://cyberbotics.com/jqueryui/1.11.4/theme.css
```
fabienrohrer: These files are also present on the CloudFlare CDN, but in a theme directory:

https://cdnjs.com/libraries/jqueryui/1.11.4

It seems our is "smoothless".

Is there some reason about not using this?

omichel: No. I thought we created our own theme, but if it is not the case, then we should use the one from the CDN.

fabienrohrer: Ok, do you want me to fix this?

omichel: Yes, please.

omichel: I guess the same applies to the images as well?

fabienrohrer: Mmm. Not sure to understand. `theme.css` is also in the `theme` folder. I included it in the fix:

https://cdnjs.cloudflare.com/ajax/libs/jqueryui/1.11.4/themes/smoothness/theme.css

PR #949, opened 2018-11-08
Attempt to fix google bot crawler parsing

omichel:
The googlebot crawler can see doc pages on this branch! 🍾
Apparently it was crashing on the `let` keyword.
I will investigate this further.
But we have a short term solution to get back in the google index!


fabienrohrer:
Hourra :-D

tn12787:
We should have had this for ages! I'm really glad we found the fix.

tn12787:
We can replace this mechanism with a static build system in the near future, but for now this should do the trick 😄

